PR #50, opened 2026-04-23
-

jonkhler:


n-gao (changes requested):
Feels oddly specific to highlight this method in particular. If we mention/use it within the notebook, a link to the API reference should suffice? :)

jonkhler:
Addressed: replaced the dedicated markdown + code + caption section with a one-sentence link to `Index.counts` inside the existing `.valid_mask` paragraph. Diff is now `+1/-1` line.

n-gao:
I don't think this is a good addition to the point being made about Buffered.

PR #56, opened 2026-04-23
docs: drop Flow from Conventions list

### LINKED ISSUE #28: Conventions chapter names Flow as a framework abstraction but the handbook never introduces it
\`docs/notebooks/conventions.md\` lists \`Flow\` alongside \`Propagator\` and \`Lens\` as examples of \"callables and framework abstractions defined by behavior\":

\`\`\`
- No verb for callables and framework abstractions defined by behavior:
  Propagator, Lens, Flow.
\`\`\`

\`Propagator\` and \`Lens\` each have their own handbook chapter. \`Flow\` does not appear anywhere else in the handbook or notebooks. The Propagators chapter builds MD steps from \`MomentumStep\`, \`PositionStep\`, \`StochasticStep\`, etc., but never mentions that position-updating propagators are parameterised by a \`Flow\`.

The protocol is a real user-facing abstraction — see \`src/kups/md/integrators.py:53\` for \`Flow\` itself and \`MinimumImageConventionFlow\` at \`src/kups/md/integrators.py:105\`. It is what carries boundary-condition behaviour (unbounded Euclidean vs. periodic minimum-image) through the integrator stack, so a reader implementing a custom integrator or building a simulation with non-standard boundaries needs to know it exists.

### Suggested fix

Two options, either is fine:

- (a) Add a short Flow subsection to \`docs/notebooks/propagators.md\`, near the \"Building an MD Step\" section where \`PositionStep\` is introduced. One paragraph on what a Flow represents (boundary-condition behaviour for position updates), a mention of \`MinimumImageConventionFlow\` as the PBC implementation, and a note that the default unbounded variant is used when boundaries are not needed.
- (b) If Flow is considered internal to MD integrators and not part of the user-facing abstraction set, drop it from the conventions.md list so the chapter does not advertise a concept the handbook never teaches.

My preference is (a): the fact that conventions.md groups it with Propagator and Lens suggests it is intended as a first-class user-facing abstraction, and the protocol signature supports that reading.

### Scope

- \`docs/notebooks/conventions.md\` (drop it) and/or \`docs/notebooks/propagators.md\` (introduce it).

jonkhler:
Fixes #28.

Reviewer n-gao pointed out that adding a Flow section to the general Propagators chapter is too MD-specific. Takes option (b) from the issue: drop `Flow` from the callable-abstractions list in `docs/notebooks/conventions.ipynb`. `Flow` remains a first-class user-facing protocol; it just doesn't get advertised in the general-audience handbook until there is a dedicated MD chapter to host a worked example.

## Test plan

- [x] `JAX_PLATFORMS=cpu uv run jupyter nbconvert --execute --to notebook docs/notebooks/conventions.ipynb` executes end-to-end.
- [x] `uv run pre-commit run --all-files` clean.

### DIFF
--- a/docs/notebooks/conventions.ipynb
+++ b/docs/notebooks/conventions.ipynb
@@ -73,7 +73,7 @@
     "- `Has*` for a single specific field or property: [HasPositions][kups.core.typing.HasPositions], [HasUnitCell][kups.core.typing.HasUnitCell], [HasSystemIndex][kups.core.typing.HasSystemIndex]. The most granular level, one protocol per field.\n",
     "- `Is*` for a multi-field role or identity: `IsState`, `IsMDState`. The object \"is\" conceptually that thing. Used when several fields always appear together.\n",
     "- `Supports*` for an operation capability: `SupportsAdd`, `SupportsDType`. Used for generic algorithms that need specific operators or methods.\n",
-    "- No verb for callables and framework abstractions defined by behavior: [Propagator][kups.core.propagator.Propagator], [Lens][kups.core.lens.Lens], `Flow`. These are already nouns that describe what the object does."
+    "- No verb for callables and framework abstractions defined by behavior: [Propagator][kups.core.propagator.Propagator], [Lens][kups.core.lens.Lens]. These are already nouns that describe what the object does."
    ]
   },
   {


n-gao (changes requested):
This feels very specific about MD. I don't think it belongs to the general discussion of propagators. I would either include this in a special MD notebook later or rely/improve the API docs :)

jonkhler:
Addressed by switching to option (b) from the issue: dropped `Flow` from the Conventions callable-abstractions list rather than adding a new section to Propagators. PR title + body updated.

n-gao (approved):

PR #62, opened 2026-04-23
-

jonkhler:


src/kups/core/propagator.py
```diff
@@ -531,6 +537,48 @@ def cond(carry):
...
+
+    Uses ``jax.lax.scan`` rather than ``jax.lax.while_loop``, so the loop is
+    reverse-mode differentiable (``jax.grad`` works end-to-end). The trip
+    count is a static Python ``int``.
+
+    Use this when you need to differentiate a scalar loss with respect to
+    parameters that flow through many propagator steps, e.g. training a
+    force-field parameter against a trajectory observable. For non-grad
+    workloads :class:`LoopPropagator` is equivalent and slightly cheaper
+    to trace.
```
n-gao: it's not true that the loop propagator is cheaper to trace, I think it's contrary. But the number of cycles does not have to be compile time static there which can be a big point of consideration.

src/kups/core/propagator.py
```diff
@@ -531,6 +537,48 @@ def cond(carry):
         return state
 
 
+@dataclass
+class ScanPropagator[State](Propagator[State]):
+    """Repeat a propagator a fixed number of times with reverse-mode AD support.
```
n-gao: I do not like the strong focus on differentiability, in this docstring. It sounds to much like "LLM addresses problem X".

docs/index.md
```diff
@@ -41,7 +41,7 @@ See the [examples/](https://github.com/cusp-ai-oss/kups/tree/main/examples) dire
 - **Molecular dynamics** — NVE, NVT, NPT ensembles
 - **Geometry optimization** — FIRE and L-BFGS relaxation
 - **Force fields** — Lennard-Jones, Coulomb (Ewald summation), harmonic bonds/angles, Morse, MACE, UMA
-- **Differentiable** — full automatic differentiation through simulations via JAX
+- **Differentiable** — automatic differentiation through simulations via JAX. Forward-mode works with any propagator; reverse-mode (`jax.grad`) requires [`ScanPropagator`](reference/kups/core/propagator.md) for loops over a static trip count.
```
n-gao: remove

n-gao (changes requested):

PR #64, opened 2026-04-23
-

jonkhler:


n-gao:
The gradients are computed with autodiff, so the only unit one needs to take care of is the energy output and that the input is expected in Angstrom, etc. Furthermore, the MD example is offtopic.

PR #80, opened 2026-04-23
-

jonkhler:


n-gao:
that's a lot of very specific extra information for the handbook's index page. I also do think the handbook should probably be a ipynb not an md file if it contains runnable code.

Issue #94, opened 2026-04-30
Batched relaxation shares one LBFGS state across independent systems, degrading convergence


samblau:
[shared_lbfgs_bug.py](https://github.com/user-attachments/files/27264118/shared_lbfgs_bug.py)

## Summary

When `relax_mlff` is given multiple independent structures via `inp_files`, all molecules share a single LBFGS optimizer state over the concatenated position vector. The L-BFGS memory vectors (`s`, `y`) mix curvature information from different potential energy surfaces, corrupting the inverse Hessian approximation and severely degrading per-system convergence.

We believe this is a promising framework and are submitting this issue (and an accompanying PR with a proposed fix) in the spirit of being helpful. This was discovered during benchmarking by Sam Blau and an AI assistant (Claude, via [Axon](https://www.mirrorphysics.com) by Mirror Physics) while evaluating kUPS for batched molecular geometry optimization with jaxified UMA.

## Reproduction

The root cause can be demonstrated with a pure-NumPy synthetic test (no kUPS or ML model needed) — see `examples/shared_lbfgs_bug.py` in the accompanying PR, and the file is also attached here. Two quadratic systems with different curvatures share one LBFGS vs having independent LBFGS instances:

```
After 50 steps:
  System A — Independent: 2.14e-11, Shared: 4.98e-02 (2327126162x worse)
  System B — Independent: 3.42e-02, Shared: 3.04e-02
```

## Impact with real ML potentials

We benchmarked 25 drug-like molecules with jaxified UMA-s-1p2 (OMol task):

| Method | Converged (fmax < 0.01 eV/Å) | Wall time |
|--------|-------------------------------|-----------|
| ASE LBFGS (sequential, per-molecule) | 22/25 | 287s |
| kUPS `relax_mlff` (shared LBFGS) | 1/25 | 145s |
| **Per-molecule vmapped LBFGS** (proposed fix) | **22/25** | **19s** |

The fix uses `jax.vmap` over per-molecule optimizer states while keeping the batched model forward pass. Step counts match ASE exactly (e.g., tafamidis: 19 steps in both).

## Root cause

In `relax_mlff.py`, the optimizer is initialized over the concatenated position vector of all systems:

```python
opt_state = opt_init(
    (particles.data.positions, systems.data.unitcell.lattice_vectors)
)
```

[...]

PR #95, opened 2026-04-30
Fix: per-system optimizer states for batched relaxation of independent systems

samblau:
### Problem

When `relax_mlff` optimizes multiple independent structures (e.g., drug molecules), all systems share a single LBFGS optimizer state over the concatenated position vector. The L-BFGS `s·y` dot products in the two-loop recursion mix curvature information from unrelated potential energy surfaces, corrupting the inverse Hessian approximation. This causes 1/25 convergence instead of 22/25.

See companion issue #94 for a minimal pure-NumPy reproduction (`examples/shared_lbfgs_bug.py`).

### Context

This was discovered by Sam Blau and an AI assistant (Claude, via [Axon](https://www.mirrorphysics.com) by Mirror Physics) while benchmarking kUPS for batched molecular geometry optimization. We think kUPS is a promising framework and hope this is useful.

### Fix

After the batched gradient computation (one forward+backward pass over all concatenated atoms), split gradients by system, pad to a common shape, and run `jax.vmap(optimizer.update)` over independent per-molecule optimizer states. The entire step — gradient + split + pad + vmapped LBFGS + unpad + scatter — compiles into a single JIT'd kernel.

### Results (25 drug molecules, UMA-s-1p2 OMol, H100)

| Method | Converged (fmax < 0.01 eV/Å) | Wall time | Per-step |
|--------|-------------------------------|-----------|----------|
| ASE LBFGS (PyTorch, sequential) | 22/25 | 287s | 105ms/mol |
| `relax_mlff` (shared LBFGS) | 1/25 | 145s | — |
| **This PR (vmapped per-molecule LBFGS)** | **22/25** | **19s** | **94ms/all 25** |

Step counts match ASE exactly (e.g., tafamidis: 19 both, dapagliflozin: 40 both, lenalidomide: 37 both). The 3 unconverged molecules are the same ones ASE can't converge in 200 steps.

### What's included

- `examples/shared_lbfgs_bug.py` — minimal pure-NumPy reproduction of the shared-LBFGS convergence bug (no kUPS or GPU needed)
- `examples/relax_mlff_vmapped.py` — standalone proof-of-concept implementing the fix using kUPS's model loading and optimizer primitives

### Design notes

This PR adds example scripts rather than modifying kUPS internals directly, since we're external contributors and wanted to demonstrate the approach without presuming how it should integrate into the Table/Lens/Propagator architecture. The key pattern is:

```python
# After batched gradient: grad shape (N_total, 3)
# Split by system using static gather indices (JIT-friendly)
padded_grad = grad_extended[gather_indices] * grad_mask  # (B, max_atoms, 3)

# Per-molecule vmapped LBFGS (independent states)
vmapped_update = jax.vmap(lambda g, s, p: optimizer.update(g, s, p))
updates, new_states = vmapped_update(padded_grad, per_mol_states, padded_pos)

# Scatter back to concatenated positions
positions = positions.at[scatter_indices].add(flat_updates)
```

A natural integration into kUPS could be a `per_system_optimizer: bool` flag in `RelaxParameters` that switches the `RelaxationPropagator` to use vmapped states. The vmapped LBFGS kernel compiles in <1s and runs in <1ms — the overhead is negligible.

### DIFF
--- a/examples/relax_mlff_vmapped.py
+++ b/examples/relax_mlff_vmapped.py
@@ -1,0 +1,445 @@
+# Copyright 2024-2026 Cusp AI
+# SPDX-License-Identifier: Apache-2.0
+
+"""Batched MLFF relaxation with per-system vmapped LBFGS optimizer states.
+
+This is a standalone proof-of-concept demonstrating the fix for the
+shared-LBFGS convergence issue in batched relaxation. It uses kUPS's
+model loading and optimizer primitives but bypasses the Table/Lens
+propagator machinery to implement the split-vmap-scatter pattern
+directly.
+
+Architecture:
+  1. Batched model forward pass: one call over all concatenated atoms
+     (kUPS-native, no padding of model inputs)
+  2. Split per-atom gradients by system using static index arrays
+  3. Pad to (B, max_atoms, 3) for vmapped optimizer update
+  4. jax.vmap(optimizer.update) — independent LBFGS state per molecule
+  5. Unpad and scatter updates back to concatenated positions
+
+The entire step (gradient + split + pad + vmap LBFGS + unpad + scatter)
+is compiled into a single JIT'd JAX kernel.
+
+Usage:
+    python relax_mlff_vmapped.py config.yaml
+
+YAML config format (same as relax_mlff, plus optional per_system_optimizer):
+    inp_files:
+      - structure1.cif
+      - structure2.xyz
+    model_path: /path/to/model.zip  # or hf://...
+    relax:
+      optimizer:
+        - transform: scale_by_ase_lbfgs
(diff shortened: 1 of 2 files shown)


n-gao:
Hi @samblau , thanks a lot for your contribution and the issue. Great work! Though for the solution I think we would like to strive for a non-vmapped solution to keep in the spirit of kUPS and composability. This most likely means that when computing vector/matrix norms in optimizers, we need to do it on a per system basis explicitly (like we do in the rest of the codebase). Using vmap to solve this problem would be incompatible with the rest of the code. 

We will look into it! :)

n-gao:
Thanks a lot! We fixed this in #109!

PR #97, opened 2026-05-04
refactor: rename UnitCell vocabulary to Cell

jonkhler:
The "unit cell" naming implied periodic crystal repetition, which does not fit the non-periodic and slab geometries. Per Nick's review, decouples the two genuinely orthogonal axes the previous design conflated, and pushes the redesign to its principled end-state.

## Design

**Frame** — geometry only.
- `OrthogonalFrame(lengths)` — 3 DOF, diagonal fast paths
- `TriclinicFrame(tril)` — 6 DOF, lower-triangular

A frame is a 3D parallelepiped. Its interpretation (periodic crystal vs bounding box vs slab) is determined by the cell that wraps it — the frame itself doesn't imply periodicity.

**Cell[P]** — frame plus per-axis boundary semantics, generic over the periodicity literal `P`.
- `PeriodicCell` — `periodic = (True, True, True)` literal. Frame is the unit cell.
- `VacuumCell` — `periodic = (False, False, False)` literal. Frame is the simulation domain's bounding parallelepiped.
- `SlabCell` — runtime per-axis mask. Frame is mixed: periodic on True axes, bounding on False.

A function declaring `Cell[FullyPeriodic]` statically rejects `VacuumCell` / `SlabCell`. Pyright preserves the literal-tuple types through the dataclass field defaults.

## Why this shape

The previous PR's single-Cell-with-runtime-mask approach used phantom marker classes (`FullyPeriodic`/`Vacuum`/`Slab` as empty types) as a Cell type parameter that the type system did not actually enforce. The new design ties periodicity into the literal type of the `periodic` field, so pyright statically discriminates.

Cell is a generic dataclass base — concrete subclasses inherit and provide a literal-typed default for `periodic`. All shared behavior (geometry passthrough, `wrap`, `__mul__`) is defined once on the base. `make_supercell` uses `dataclasses.replace` to rebuild the same concrete cell type with a tiled frame; no isinstance ladder, no type-ignore.

The `Lattice` name was abandoned because it reads dishonestly for vacuum cells: a vacuum simulation has no periodic crystal structure, only a bounding box. `Frame` is geometry-agnostic to the periodicity question and reads cleanly across all three cell types.

The `L` (lattice/frame) type parameter was dropped from `Cell` and `HasCell`. Zero production callsites narrowed on the lattice axis — the lattice-vs-frame discrimination was speculative future-proofing. Frame-specific narrowing (`OrthogonalFrame.lengths`, `TriclinicFrame.tril`) is now via inline `isinstance(cell.frame, ...)`. Annotations are correspondingly less noisy: `Cell[FullyPeriodic]` instead of `Cell[Lattice, FullyPeriodic]`.

## Type-narrowing landed during this redesign

- `IsIdealGasSystems(HasTemperature, HasCell[FullyPeriodic])` — `N·k_B·T/V` requires PBC.
- `IsVirialSystems(HasCell[FullyPeriodic], Protocol)` — virial stress requires PBC.
- RDF `cell: Cell[FullyPeriodic]` — RDF is a periodic concept.

These narrowings were left on the table by the initial split; the new design exists to enable exactly these.

## Concrete impact

- `cell.py` reduced ~125 lines (de-triplication on the cell subclasses + collapsed `make_supercell`).
- One `# pyright: ignore` removed.
- 44 files touched in this commit (10 src + 30 test + 3 notebooks + 1 typing).
- 932 tests pass.
- All semantically-stale `unit_cell`/`uc` vocabulary scrubbed.

## Renames in this redesign (one-shot, no shims)

- `Lattice` → `Frame`
- `OrthogonalLattice` → `OrthogonalFrame`
- `TriclinicLattice` → `TriclinicFrame`
- `cell.lattice` → `cell.frame`
- `.lattice_vectors` → `.vectors`
- `.inverse_lattice_vectors` → `.inverse_vectors`
- `Lattice.scale(multiplicities)` → `Frame.tile(multiplicities)` (matches `numpy.tile` semantics)
- `Cell[L: Lattice, P: tuple[bool, bool, bool]]` → `Cell[P: tuple[bool, bool, bool]]`
- `HasCell[L: Lattice, P: tuple[bool, bool, bool]]` → `HasCell[P: tuple[bool, bool, bool]]`


n-gao (changes requested):
Instead of having empty classes to indicate whether something is Vacuum or periodic, etc. We could decouple the representation of the lattice vectors (orthogonal, tricilinic) from the periodicity. Periodicity/BoundaryCondition/Cell would be a wrapper around the Lattice vectors with subclasses that return literals for periodicity. This way we decouple the representation of the lattice vectors while still allowing for static type checks on the periodicity.

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+
+"""Simulation cell representations.
+
+A cell wraps a [Frame][kups.core.cell.Frame] (a 3D parallelepiped) and adds
+a per-axis periodicity mask. The frame itself is geometry only — the same
+frame can describe a periodic crystal's unit cell, a vacuum simulation's
+bounding box, or the bounding parallelepiped of a slab. Which interpretation
+applies is decided by the cell type that wraps the frame.
+
+## Frame — geometry
```
n-gao: Lattice/geometry? We can also keep Frame but without context I wouldn't directly think of lattice vectors there.

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+@dataclass
+class PeriodicCell(Cell[FullyPeriodic]):
+    """Cell that is periodic along all three axes.
+
+    The ``periodic`` field is pinned to the literal type ``FullyPeriodic``,
+    so a function declared ``Cell[FullyPeriodic]`` accepts ``PeriodicCell``
+    and rejects ``VacuumCell`` / ``SlabCell`` statically.
+    """
+
+    periodic: FullyPeriodic = field(default=(True, True, True), static=True)
```
n-gao: should be a property to be read only

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+@dataclass
+class VacuumCell(Cell[Vacuum]):
+    """Cell with all three axes open. The frame is the simulation domain's
+    bounding parallelepiped — used by neighbor-list spatial partitioning
+    and any consumer that needs a domain size.
+    """
+
+    periodic: Vacuum = field(default=(False, False, False), static=True)
```
n-gao: should be a property to be read only

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+    """Frame plus per-axis boundary semantics.
+
+    Generic over the periodicity literal ``P``. Three concrete subclasses
+    pin ``P`` to a literal-typed tuple (PeriodicCell, VacuumCell) or leave
+    it as a runtime mask (SlabCell). All shared behavior — geometry
+    passthrough, wrap, scaling — lives on this base.
+    """
+
+    frame: Frame
+    periodic: P = field(static=True)
```
n-gao: should be a property to be read only and then overwritten by the others?

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+        FRACTIONAL: Scaled coordinates in [0, 1) relative to frame vectors.
+    """
+
+    REAL = "real"
+    FRACTIONAL = "fractional"
+
+
+type FullyPeriodic = tuple[Literal[True], Literal[True], Literal[True]]
+type Vacuum = tuple[Literal[False], Literal[False], Literal[False]]
+type AnyMask = tuple[bool, bool, bool]
```
n-gao: AnyPeriodicity

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+    Attributes:
+        REAL: Cartesian coordinates in Angstroms.
+        FRACTIONAL: Scaled coordinates in [0, 1) relative to frame vectors.
+    """
+
+    REAL = "real"
+    FRACTIONAL = "fractional"
+
+
+type FullyPeriodic = tuple[Literal[True], Literal[True], Literal[True]]
```
n-gao: Periodic3D

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+) -> TypeGuard[VacuumCell]:
+    """``True`` iff ``cell`` is a [VacuumCell][kups.core.cell.VacuumCell]."""
+    return isinstance(cell, VacuumCell)
+
+
+def is_fully_periodic[P: tuple[bool, bool, bool]](
+    cell: Cell[P],
+) -> TypeGuard[PeriodicCell]:
+    """``True`` iff ``cell`` is a [PeriodicCell][kups.core.cell.PeriodicCell]."""
+    return isinstance(cell, PeriodicCell)
```
n-gao: A general Cell can also be fully periodic, I would check all(cell.periodic).

I would also suggest a more explicit name like `is_3d_periodic`

src/kups/core/cell.py
```diff
@@ -0,0 +1,472 @@
...
+    """Cell with all three axes open. The frame is the simulation domain's
+    bounding parallelepiped — used by neighbor-list spatial partitioning
+    and any consumer that needs a domain size.
+    """
+
+    periodic: Vacuum = field(default=(False, False, False), static=True)
+
+
+@dataclass
+class SlabCell(Cell[AnyMask]):
```
n-gao: Isn't a slab strictly 2D periodic? A vacuum or 3d lattice would not be considered a slab?

n-gao (changes requested):
Minor suggestions regarding strict type checking

src/kups/core/cell.py
n-gao: another option would be to have `init=False` in all subclasses! :)

src/kups/core/cell.py
n-gao: Not sure whether we weed this class to begin with if it doesn't enforce anything new?

src/kups/core/cell.py
n-gao: alternative: `init=False` in `field`

src/kups/core/cell.py
n-gao: alternative: `init=False` in `field`

src/kups/core/cell.py
jonkhler: Lattice is explicitly about repeating things - frame is more faithful of what we are talking about - basically a bounding box in the non repeating case. I would prefer a precise term based on what it is about rather domain-adjacent but wrongly coined ones. Frame captures nicely what it is. But happy about other recommendations.

src/kups/core/cell.py
jonkhler: agreed

src/kups/core/cell.py
jonkhler: agreed

src/kups/core/cell.py
jonkhler: agreed

src/kups/core/cell.py
jonkhler: agreed

src/kups/core/cell.py
jonkhler: agreed

src/kups/core/cell.py
jonkhler: we can also remove it for now!

src/kups/core/cell.py
jonkhler: ok agree!

n-gao (approved):
LGTM

Issue #89, opened 2026-04-24
Support non-periodic systems in `CellListNeighborList`


sirmarcel:
### Problem or Use Case

Hey everyone, nice project. A small request:

It would be nice if `CellListNeighborList` would support not only periodic systems, but also non-periodic cases (properly prepared so all positions are guaranteed to be inside some bounding box). This unlocks both large molecules and things like the [unfolded heat flux](https://github.com/sirmarcel/glp/blob/main/glp/calculators/heat_flux_unfolded.py) that is needed to compute thermal conductivity with MPNNs and the Green-Kubo method. (As a test of `kUPS` I'm currently wiring up my old PhD code for this workflow.)

Not sure if non-pbc is at all in-domain for this project, so feel free to close! I'll likely hack something together for the fun of it, but it might be nice to support properly eventually.

### Proposed Solution

I think this requires (a) guarding for positions inside the bounding box (since you can't wrap for non-pbc) and (b) adapting the filtering of candidates after the cell list pre-selections step to exclude cross-surface pairs.

This wastes some compute (you'd still initially suggest candidates across boundaries and later throw them away) but it's probably worth checking if the complexity of adapting the selection is worth the payoff later. After all, in the big system limit, where the cell list is most useful, the surface terms should be small.

### Alternatives Considered

The fallback is using the $N^2$ neighbourlist. This is fine, and likely not terribly slow for most systems. However, the final benchmark in this little side project would be trying to converge the thermal conductivity of Si, which needs boxes >10k atoms and is likely to hit the limits of the naive approach.

Another alternative is making a non-pbc system with a `> cutoff` gap between all atoms and the cell surface. This is fairly wasteful (but probably also needs benchmarking to exclude for certain). In a really big box, the waste is maybe also small...

jonkhler:
Thanks for that issue @sirmarcel! 

I spent some time today and I think I found a good solution that should support that. I will try to get a polished PR up by Monday - then we could have a look together. But on first sight (a) requires some API change with the neighborlist that would make things a lot simpler/cleaner now and (b) the heavylifting is very straight forward then. So I guess this will be something like two PRs. Will ping you when online!

ArpBansal:
@jonkhler you worked on this issue? If not should i try?
If i felt for changes in api, will first discuss the changes in here itself first, only then will think about moving forward.

jonkhler:
[...]

PR #98, opened 2026-05-04
feat: Implement per-axis periodic mask support in cell list

jonkhler:
Implements the non-PBC story end-to-end on top of the new Frame/Cell hierarchy: the cell list (and the shared `basic_neighborlist` path) now honor the cell's `periodic` tuple so that `VacuumCell` produces correct edges, and the ASE loader propagates `atoms.pbc` into the kups cell so simulations driven through the application layer use the right boundary mode. Unblocks Marcel's request (issue #89) for non-PBC support.

## Three masked sites in `core/neighborlist.py`

- `_compute_distances_pbc_sq`: MIC fold gated as `jnp.where(periodic, jnp.round(deltas), 0.0)` so non-periodic axes get a zero shift.
- `_get_candidate_images`: image counts on open axes forced to 1 so the existing early-out fires for fully-open systems.
- `_cell_list_subselect`: trace-time gate on `all(periodic)`. The fully-periodic branch is byte-identical to the prior code (verified via HLO inspection); the general branch wraps only on periodic axes and routes out-of-range stencil offsets to the `cell_oob` sentinel so cross-boundary candidates produce no key matches.

## Application loader

`particles_from_ase` reads `atoms.pbc` and constructs the cell accordingly:
- all-True → `PeriodicCell(TriclinicFrame.from_matrix(...))`
- all-False → `VacuumCell(TriclinicFrame.from_matrix(...))`
- mixed → `raise NotImplementedError(...)` with a clear message

Mixed-axis (slab / wire) PBC raises today because the cell hierarchy currently exposes only the all-True/all-False literals. The neighbor-list machinery is already mask-aware via `cell.periodic` and will work transparently once a slab cell type is re-introduced (no neighborlist-side changes will be needed).

## Tests in `TestNonPeriodicNeighborList`

- Cross-boundary pair excluded under `VacuumCell`.
- In-box pair found.
- `CellListNeighborList` and `DenseNearestNeighborList` agree on a non-PBC fixture (locks in that the cell-list-specific stencil routing is consistent with the dense reference, since both go through `basic_neighborlist`).
- `PeriodicCell` path unchanged (sanity).

## Docstring updates

`core/neighborlist.py`, `core/utils/position.py`, `observables/radial_distribution_function.py`, and `potential/common/graph.py` replace "requires/uses periodic boundary conditions" with the per-axis story, since the entire stack now reads `cell.periodic` and works correctly for vacuum and bulk.

### DIFF
--- a/src/kups/application/utils/particles.py
+++ b/src/kups/application/utils/particles.py
@@ -13,7 +13,13 @@ import ase.io
 import jax.numpy as jnp
 from jax import Array
 
-from kups.core.cell import Cell, PeriodicCell, TriclinicFrame, to_lower_triangular
+from kups.core.cell import (
+    Cell,
+    PeriodicCell,
+    TriclinicFrame,
+    VacuumCell,
+    to_lower_triangular,
+)
 from kups.core.data import Index, Table
 from kups.core.typing import ExclusionId, InclusionId, Label, ParticleId, SystemId
 from kups.core.utils.jax import dataclass
@@ -73,7 +79,24 @@ def particles_from_ase(
     if isinstance(atoms, (str, Path)):
         atoms = next(ase.io.iread(atoms, index=-1, store_tags=True))
     L, uc_transform = to_lower_triangular(jnp.asarray(atoms.cell.array))
-    cell = PeriodicCell(TriclinicFrame.from_matrix(L))
+    pbc: tuple[bool, bool, bool] = (
+        bool(atoms.pbc[0]),
+        bool(atoms.pbc[1]),
+        bool(atoms.pbc[2]),
+    )
+    cell: Cell
+    if all(pbc):
+        cell = PeriodicCell(TriclinicFrame.from_matrix(L))
+    elif not any(pbc):
+        cell = VacuumCell(TriclinicFrame.from_matrix(L))
+    else:
+        # Mixed-axis periodicity (slab / wire) needs a runtime per-axis mask;
+        # the current Cell hierarchy only exposes the all-True / all-False
+        # literals. Reintroduce a slab-shaped cell when a real consumer lands.
+        raise NotImplementedError(
+            f"Mixed per-axis periodicity {pbc} is not supported. "
+            "Use uniform pbc=True (PeriodicCell) or pbc=False (VacuumCell)."
+        )
     # Rotate Cartesian positions into the lower-triangular frame.
     positions = uc_transform(jnp.asarray(atoms.positions))
     masses = jnp.asarray(atoms.get_masses())
--- a/src/kups/application/utils/particles.py
+++ b/src/kups/application/utils/particles.py
@@ -13,13 +13,7 @@ import ase.io
 import jax.numpy as jnp
 from jax import Array
 
-from kups.core.cell import (
-    Cell,
-    PeriodicCell,
-    TriclinicFrame,
-    VacuumCell,
-    to_lower_triangular,
-)
+from kups.core.cell import Cell, TriclinicFrame, to_lower_triangular
 from kups.core.data import Index, Table
 from kups.core.typing import ExclusionId, InclusionId, Label, ParticleId, SystemId
 from kups.core.utils.jax import dataclass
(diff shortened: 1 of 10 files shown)


src/kups/application/utils/particles.py
```diff
@@ -73,7 +79,24 @@ def particles_from_ase(
...
+        bool(atoms.pbc[0]),
+        bool(atoms.pbc[1]),
+        bool(atoms.pbc[2]),
+    )
+    cell: Cell
+    if all(pbc):
+        cell = PeriodicCell(TriclinicFrame.from_matrix(L))
+    elif not any(pbc):
+        cell = VacuumCell(TriclinicFrame.from_matrix(L))
+    else:
```
n-gao: Couldn't we instantiate a Cell here?

src/kups/core/neighborlist.py
```diff
@@ -298,8 +299,20 @@ def _cell_list_subselect(
...
+        key_positions = _wrap_coordinates(lh.data.positions)
+        query_positions = _wrap_coordinates(rh.data.positions)
+    else:
+        mask = jnp.array(periodic)
+        key_positions = jnp.where(
+            mask, _wrap_coordinates(lh.data.positions), lh.data.positions
+        )
+        query_positions = jnp.where(
+            mask, _wrap_coordinates(rh.data.positions), rh.data.positions
+        )
```
n-gao: Shouldn't cell's .wrap take care of this?

src/kups/core/neighborlist.py
```diff
@@ -325,17 +338,31 @@ def _cell_list_subselect(
...
+        # offsets on open axes route to cell_oob so they produce no key matches
+        # (cross-boundary candidates are excluded).
+        mask = jnp.array(periodic)
+        shifted = jnp.where(mask, _wrap_coordinates(raw_shifted), raw_shifted)
+        in_box = jnp.all((shifted >= 0) & (shifted < 1), axis=-1)
+        hashes = (
+            _cell_hash(shifted, num_cells[query_system])
+            + rh_system_ids[query_original.indices] * max_num_cells.size
+        )
+        query_neighborhood_hashes = jnp.where(in_box, hashes, cell_oob)
```
n-gao: Again cell.wrap should handle the logic of periodic. Ideally, we do not need to concern ourselfs with .periodic outside as much as possible

src/kups/core/neighborlist.py
```diff
@@ -553,7 +583,8 @@ def _compute_distances_pbc_sq(
         - rh.data.positions[candidates.rhs.indices]
     )
     if shifts is None:
-        shifts = jnp.round(deltas)
+        periodic = jnp.array(systems.data.cell.periodic)
+        shifts = jnp.where(periodic, jnp.round(deltas), 0.0)
```
n-gao: Should be a Cell method

test/core/test_neighborlist.py
```diff
@@ -1762,3 +1768,106 @@ def test_random_large(self):
...
+        avg_image_candidates=FixedCapacity(max(n * n, 8)),
+    )
+    result = jax.jit(as_result_function(nl))(
+        lh=lh, rh=None, systems=systems, cutoffs=cutoffs, rh_index_remap=None
+    )
+    result.raise_assertion()
+    return result.value
+
+
+class TestNonPeriodicNeighborList:
```
n-gao: Can we directly test all variations of periodicities? vaccuum, 1D, 2D, 3d?

src/kups/core/utils/position.py
n-gao, line 1: This file needs to be updated as well to respect periodicity, right?

n-gao (changes requested):
The general changes are sound to me! However, I would avoid the user-side code (in this case everything outside of cell.py) dealing with the .periodic tuple as much as possible and would instead encapsulate most of the handling to the existing .wrap method and new methods were appropriate. This would help us to reduce errors in the future.

src/kups/core/cell.py
```diff
@@ -465,9 +495,7 @@ class PeriodicCell(Cell[Periodic3D]):
     Ewald summation requires this cell type.
     """
 
-    @property
-    def periodic(self) -> Periodic3D:
-        return (True, True, True)
+    periodic: Periodic3D = field(default=(True, True, True), static=True)
```
n-gao: init=False

src/kups/core/cell.py
```diff
@@ -481,9 +509,7 @@ class VacuumCell(Cell[Vacuum]):
     neighbor-list machinery needs a spatial-partitioning hint.
     """
 
-    @property
-    def periodic(self) -> Vacuum:
-        return (False, False, False)
+    periodic: Vacuum = field(default=(False, False, False), static=True)
```
n-gao: init=False

src/kups/application/utils/particles.py
```diff
@@ -73,8 +79,22 @@ def particles_from_ase(
...
+    cell: Cell
+    match pbc:
+        case (True, True, True):
+            cell = PeriodicCell(frame)
+        case (False, False, False):
+            cell = VacuumCell(frame)
+        case (True, True, False) | (True, False, True) | (False, True, True):
+            cell = Cell(frame, periodic=pbc)
+        case (True, False, False) | (False, True, False) | (False, False, True):
+            cell = Cell(frame, periodic=pbc)
```
n-gao: this could be a staticmethod of Cell?

n-gao (approved):
LGTM, we could introduce a staticmethod for constructing the cell and set init=False for the Vacuum and PeriodicCell.

PR #113, opened 2026-05-12
chore: Bump version to 1.0.1

n-gao:
> 🤖 *Auto-generated by Claude. Review and edit as needed.*

Bump version from 1.0.0 to 1.0.1 in pyproject.toml.

PR #114, opened 2026-05-12
fix: carry init=False static fields in pytree aux_data

jonkhler:
## Summary

- `PeriodicCell.periodic` and `VacuumCell.periodic` are declared `init=False, static=True`. The kups `dataclass` utility partitioned them into JAX's `drop_fields`, removing them from the pytree's aux_data. The resulting treedefs satisfy `td_v == td_p` (reprs differ; hashes differ; equality does not), violating the hash/eq contract and leaving a sharp edge where a treedef-keyed JIT cache can mis-route a vacuum call onto a previously-traced periodic compilation.
- For dataclasses with any `init=False AND static=True` field, switch from `register_dataclass(drop_fields=...)` to `register_pytree_node` with custom flatten/unflatten that carries those values in the aux_data. Unflatten still skips them when calling `__init__` — constructor contract unchanged.
- Adds `test_sibling_subclasses_with_distinct_pins_have_distinct_treedefs`, `test_treedef_keyed_cache_distinguishes_pins_under_hash_collision`, and `lax.cond` / `lax.scan` regressions.

## Validation (Linux x86_64, Python 3.12.3, `PYTHONHASHSEED=12`, single-process, fixed seed)

Reran the failing class 366× without the fix and 365× with the fix:

| | Baseline (no fix) | With this PR |
|---|---|---|
| Failures | 10 / 366 (2.7%) | 0 / 365 |
| Wilson 95% CI | [1.5%, 5.0%] | [0%, 1.0%] |

- Fisher's exact one-tailed `p = 9.3 × 10⁻⁴`
- Failure signature matches CI exactly: `assert {(0, 1), (1, 0)} == set()` — Vacuum producing the Periodic neighborlist
- Could not reproduce on Mac aarch64 (35+ runs); reproduces readily on Linux x86_64

## Mechanism caveat

The exact JAX-internal path that turns the eq violation into a ~3% mis-route is not fully isolated:
- `hash(td_p) == hash(td_v)` collision rate across 100 fresh processes: **0** — so it is *not* a dict-keyed JIT cache hash collision as originally hypothesized
- Flake persists when bypassing `as_result_function` (the `lax.cond` wrapper) — so it's not the user-code `lax.cond` path
- Most likely candidate: a JAX C++ cache fast-path that consults eq before / instead of hash, or object-id reuse after GC

The fix is correct regardless: it removes the JAX hash/eq contract violation and empirically eliminates the flake.

## Test plan

- [x] `pytest test/core/test_jax_utils.py::TestDataclassSubclassPinsInheritedField` — new regressions pass
- [x] `pytest test/core/test_neighborlist.py::TestNeighborListAcrossPeriodicities` — 0/365 fails with fix
- [x] Full suite — no regressions
- [x] `pyright src/kups/core/utils/jax.py test/core/test_jax_utils.py` — 0 errors

### DIFF
--- a/src/kups/core/utils/jax.py
+++ b/src/kups/core/utils/jax.py
@@ -767,9 +767,7 @@ def dataclass[T: type](
             kwargs.update((n, v) for n, v in zip(_meta, meta) if n in _init_meta)
             return _cls(**kwargs)
 
-        jax.tree_util.register_pytree_with_keys(
-            dcls, _flatten_with_keys, _unflatten
-        )
+        jax.tree_util.register_pytree_with_keys(dcls, _flatten_with_keys, _unflatten)
         return dcls
 
     if cls is None:
--- a/src/kups/core/utils/jax.py
+++ b/src/kups/core/utils/jax.py
@@ -725,66 +725,51 @@ def dataclass[T: type](
             weakref_slot=weakref_slot,
         )  # type: ignore
         dcls._jax_dataclass = True
-        # Partition fields for pytree registration. `init=False` fields cannot
-        # be passed to the constructor, so they go to JAX's `drop_fields` —
-        # except when they are also `static=True`, in which case their value is
-        # a class-level constant that distinguishes one subclass from another
-        # (e.g. `PeriodicCell.periodic == (True, True, True)` versus
-        # `VacuumCell.periodic == (False, False, False)`). Such fields must be
-        # carried in the pytree's aux_data so the resulting treedef hashes and
-        # compares equal only for cells of the same flavor; otherwise JAX
-        # treats sibling subclasses as treedef-equal and the JIT cache can
-        # mis-route a vacuum call onto a previously-traced periodic
-        # compilation.
-        init_fields = [f for f in dataclasses.fields(dcls) if f.init]
-        non_init_fields = [f for f in dataclasses.fields(dcls) if not f.init]
+        # All static fields (whether `init=True` or `init=False`) are carried
+        # in the pytree's aux_data, so the resulting treedef hashes and
+        # compares equal only for instances whose static metadata agrees on
+        # every field. This matters in particular for the
+        # `PeriodicCell.periodic == (True, True, True)` /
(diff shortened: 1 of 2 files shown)


n-gao (changes requested):
Oh I see the reason for the flaky test was that it depends on the order of execution? I.e., since two variants get the same hashes, the jit cache lookup resulted in the same function?

If this is true, it would be great to capture this in a test to be sure. The current test fails on main. Though, in isolation I cannot really tell that it is what caused the flaky CI.

jonkhler:
Added a second test that forces a hash collision so the dict falls back to eq — len(cache) == 1 on main, 2 with the fix. Both new tests fail on main.

jonkhler:
Fair point — dropped the synthetic Colliding test. I tried to build a real-jit regression (loop over PYTHONHASHSEEDs, alternating subclasses through jit, looking for any wrong result) but couldn't make it fail on Mac in 30 × 200 iterations. The jit cache uses hash-first dict lookup, hashes differ, so eq is never compared in practice.

So the test now only asserts the concrete invariant: `td_a != td_b` for sibling subclasses with distinct pinned defaults. Fails on main, passes with the fix. The jit-cache-mis-routing claim was my hypothesis for the CI flake but isn't proven — I've reframed the PR to say so.

jonkhler:
Updated with cond + scan tests that demonstrate the bug through real JAX HOFs (no synthetic dict). Both silently return the wrong subclass on main; with the fix they raise `TypeError` from JAX's branch-structure check.

Plain jit isn't affected — its cache uses hash-first dict lookup, hashes differ, so vacuum and periodic don't collide there. Dropped a jit-only test because it would pass on both main and fix.

Also dropped the framing that this fixes the CI flake on #113. The failing tests there are all-vacuum, no mixed-subclass branches, so the cond/scan bug can't fire on that code path. Flake remains unexplained — I'll keep digging on the cell-list candidate generation separately.

n-gao (approved):
LGTM

PR #126, opened 2026-05-19
feat: Widom test-particle insertion

jonkhler:
- `kups.mcmc.widom`: `widom_test`, `GhostProbe`, `WidomStatistics`, `finalize_widom` (Vlugt 2008 eq. 16, N=0).
- `kups_mcmc_widom` CLI: NVT chain + Widom probes per cycle. Auto-includes Ewald / blocking spheres from state.
- `examples/lennard_jones/uff4mof.yaml`: UFF + UFF4MOF + UFF4MOF-II + TraPPE sites, transcribed from lammps-interface (BSD).
- `examples/mcmc_widom.yaml`: RUBTAK + CO2 default.
- `WidomHostConfig.atom_type_map`: optional CIF-label rewrite for FF re-routing.
- `ExchangeMove._propose_insertion` -> `propose_insertion` (public; reused by the Widom probe).

Smoke (RUBTAK + CO2 at 1 bar / 298 K, 500 cycles): mu_ex = -0.145 eV, K_H = 6.66e8 A^3/eV, q_st = 0.266 eV (25.6 kJ/mol).

## Test plan
- [x] `pytest test/mcmc/ test/application/test_mcmc_widom.py test/application/test_mcmc_rigid.py`: 70 passed.
- [x] `pyright`: 0 errors.
- [x] `kups_mcmc_widom examples/mcmc_widom.yaml` end-to-end.


src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,224 @@
...
+            n_samples=jnp.zeros(n_systems, dtype=jnp.int32),
+        )
+
+    def reset(self) -> WidomStatistics:
+        """Zero all fields while preserving shape/dtype."""
+        return WidomStatistics(
+            sum_boltzmann=jnp.zeros_like(self.sum_boltzmann),
+            sum_delta_u_boltzmann=jnp.zeros_like(self.sum_delta_u_boltzmann),
+            n_samples=jnp.zeros_like(self.n_samples),
+        )
```
felixmusil:
```suggestion
        return self.zeros(int(self.n_samples.shape[0]))
```

src/kups/application/simulations/mcmc_widom.py
```diff
@@ -0,0 +1,385 @@
...
+        patch_fn=MCMCStateUpdate.from_changes,
+        log_probability_ratio_fn=boltzmann_ratio,
+        stat_lens=lens(lambda s: s.widom_statistics.data, cls=WidomState),
+        update_fn=_update_widom_stats,
+    )
+    widom_loop: Propagator[WidomState] = LoopPropagator(
+        widom_probe, config.num_widom_per_cycle
+    )
+
+    production = SequentialPropagator((ResetOnErrorPropagator(nvt_loop), widom_loop))
```
n-gao: The `ResetOnErrorPropagator` should wrap everything to avoid collecting widom statistics on iterations where assertions fail

src/kups/application/simulations/mcmc_widom.py
```diff
@@ -0,0 +1,385 @@
...
+    # a naked Python loop skips the donation and leaks O(state size) per cycle.
+    state = run_warmup_cycles(
+        next(chain), propagator, state, config.run.num_warmup_cycles
+    )
+    state = replace(
+        state,
+        widom_statistics=Table(
+            state.widom_statistics.keys, state.widom_statistics.data.reset()
+        ),
+    )
```
n-gao, line 354: use lenses instead? ;-) `bind(state, lambda x: x.widom_statistics.data).apply(WidomStatistics.reset)`

src/kups/application/simulations/mcmc_widom.py
n-gao, line 1: Here all the statistics are aggregated within the state, why not log it to disk and analyze post-hoc?

src/kups/application/simulations/mcmc_widom.py
```diff
@@ -0,0 +1,385 @@
...
+        cell=state_lens.focus(lambda x: x.systems.map_data(lambda d: d.cell)),
+        capacity=state_lens.focus(lambda x: x.move_capacity),
+    )
+    widom_probe = GhostProbe(
+        propose_fn=exchange.propose_insertion,
+        patch_fn=MCMCStateUpdate.from_changes,
+        log_probability_ratio_fn=boltzmann_ratio,
+        stat_lens=lens(lambda s: s.widom_statistics.data, cls=WidomState),
+        update_fn=_update_widom_stats,
+    )
```
n-gao: can we add a handy constructor from state like the other interfaces to bundle this wireing?

docs/simulations.md
```diff
@@ -72,6 +72,23 @@ kups_mcmc_rigid mcmc_rigid.yaml
...
+| `kups_mcmc_widom` | Lennard-Jones + Ewald | Widom ghost insertion for $\mu^\mathrm{ex}$, $K_H$, $q_\mathrm{st}$ |
+
+```sh
+cd examples
+kups_mcmc_widom mcmc_widom.yaml
+```
+
+[`finalize_widom`][kups.mcmc.widom.finalize_widom] applied to the accumulated [`WidomStatistics`][kups.mcmc.widom.WidomStatistics] yields $\mu^\mathrm{ex}$, $K_H$, and $q_\mathrm{st}$ (Vlugt 2008 eq. 16, $N=0$). Multiple hosts in one config become parallel batched systems on the `Table[SystemId, ...]` axis; duplicate a host entry $k$ times for $k$ independent chains. Blocking spheres ([`HostConfig.blocking_spheres`][kups.application.mcmc.data.HostConfig.blocking_spheres]) exclude regions from insertion.
+
+The default config uses [UFF4MOF / UFF4MOF-II](https://github.com/cusp-ai-oss/kups/blob/main/examples/lennard_jones/uff4mof.yaml) with an `atom_type_map:` rewriting the bare-element labels in `RUBTAK.cif` to UFF4MOF type names at host load. A vanilla-UFF (Rappé / `trappe.yaml`) fallback sits commented next to the active import.
```
n-gao: I am not sure how helpful this text is.

src/kups/application/simulations/mcmc_widom.py
```diff
@@ -0,0 +1,385 @@
...
+class WidomHostConfig(HostConfig):
+    """``HostConfig`` with an optional CIF-label rename."""
+
+    atom_type_map: dict[str, str] | None = None
+    """Optional CIF-label rename applied at host load.
+
+    E.g. ``{"Zr": "Zr8f4", "C": "C_R", "O": "O_R", "H": "H_"}`` rewrites
+    bare-element labels to UFF4MOF type names so the LJ lookup matches
+    ``lennard_jones/uff4mof.yaml``. Labels absent from the map pass through.
+    """
```
n-gao: This is not how we should handle this! This is quite tiresome setup, we should just use the same labels in the FF parameters as in the cif file, as we do for the other lennard jones examples..

test/application/test_mcmc_widom.py
```diff
@@ -0,0 +1,202 @@
...
+
+
+@pytest.fixture(autouse=True, scope="module")
+def clear_cache():
+    jax.clear_caches()
+    gc.collect()
+    yield
+    jax.clear_caches()
+    gc.collect()
+
```
n-gao: this can be imported

test/mcmc/test_widom.py
```diff
@@ -0,0 +1,223 @@
...
+)
+from kups.core.typing import SystemId
+from kups.core.utils.jax import dataclass
+from kups.mcmc.widom import (
+    GhostProbe,
+    WidomStatistics,
+    finalize_widom,
+    widom_test,
+)
+
```
n-gao, line 29: cache_clear should be imported

n-gao (changes requested):
A few comments on code consistency with the rest of the code base :)

src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,224 @@
...
+        sum_delta_u_boltzmann:
+            $\sum \Delta U \cdot \exp(-\beta \Delta U)$ [energy], with
+            $\Delta U$ the ghost insertion (host-guest) energy, not the
+            cell's total potential energy.
+        n_samples: Number of evaluations accumulated.
+    """
+
+    sum_boltzmann: Array
+    sum_delta_u_boltzmann: Array
+    n_samples: Array
```
felixmusil, line 102: I think we need to store the elements of the sum. Not only the average is important but also the estimate of the error in the mean (\epsilon) for which we need 

Image

Image

 with M the number of independent samples, and \tau_A is the estimated autocorrelation time. You can have a look [here](https://github.com/Cusp-AI/platform/blob/19847cb76455bf4bd581f782fcf0764f0bfb61ba/libraries/cuspy_chemistry_utils/src/cuspy_chemistry_utils/md/average.py#L25)  for such analysis in the context of MD.

src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,224 @@
...
+            $q_\mathrm{st} = k_BT
+            - \langle \Delta U \cdot \exp(-\beta \Delta U)\rangle / \langle \exp(-\beta \Delta U)\rangle$
+            [energy], with intramolecular gas energy taken as zero (rigid
+            adsorbate).
+    """
+
+    excess_chemical_potential: Energy
+    henry_coefficient: Array
+    heat_of_adsorption: Energy
+
```
felixmusil: same comment as above on the need for statistical error estimates here.

src/kups/mcmc/widom.py
felixmusil: In this particular case since we don't really have a chain, the error in the mean should simplify into `\sigma/\sqrt[M]`

src/kups/application/mcmc/analysis.py
```diff
@@ -217,6 +222,143 @@ def analyze_mcmc(
...
+    zero_du = jnp.zeros_like(sum_du_w[:1])
+    zero_n = jnp.zeros_like(n[:1])
+    delta_w = jnp.diff(jnp.concatenate([zero_w, sum_w], axis=0), axis=0)
+    delta_du_w = jnp.diff(jnp.concatenate([zero_du, sum_du_w], axis=0), axis=0)
+    delta_n = jnp.diff(jnp.concatenate([zero_n, n], axis=0), axis=0)
+    return delta_w / delta_n, delta_du_w / delta_n
+
+
+@no_jax_tracing
+def _analyze_single_widom_system(
```
n-gao, line 263: if we don't allow for jit tracing, we should use numpy not jax for computations.

n-gao (changes requested):

src/kups/application/mcmc/analysis.py
jonkhler: *cough* *cough*

src/kups/mcmc/moves.py
n-gao: Out of symmetry, this one should probably be public as well?

src/kups/mcmc/moves.py
```diff
@@ -800,7 +800,7 @@ def _zero_ratio(self, state: State) -> LogProbabilityRatio:
         n_sys = self.positions(state).data.system.num_labels
         return Table.arange(jnp.zeros((n_sys,)), label=SystemId)
 
-    def _propose_insertion(
+    def propose_insertion(
```
n-gao, line 803: public methods should have a docstring

src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,220 @@
...
+    kT = temperature * BOLTZMANN_CONSTANT
+    # Vlugt 2008 eq. 16 (N=0): W-weighted mean of ΔU. No covariance term;
+    # the cell's total PE does not enter. With a constant host PE a
+    # covariance form would collapse q_st to kT.
+    return WidomResult(
+        excess_chemical_potential=-kT * jnp.log(mean_w),
+        henry_coefficient=volume * mean_w / kT,
+        heat_of_adsorption=kT - mean_du_w / mean_w,
+    )
+
```
n-gao: This method is referenced a lot in docstrings but never used?

src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,220 @@
...
+        heat_of_adsorption: Vlugt (2008) eq. 16, test-particle ($N = 0$) limit
+            $q_\mathrm{st} = k_BT
+            - \langle \Delta U \cdot \exp(-\beta \Delta U)\rangle / \langle \exp(-\beta \Delta U)\rangle$
+            [energy], with intramolecular gas energy taken as zero (rigid
+            adsorbate).
+    """
+
+    excess_chemical_potential: Energy
+    henry_coefficient: Array
+    heat_of_adsorption: Energy
```
n-gao: This is never used. (since finalize_widom is never used)

docs/simulations.md
```diff
@@ -72,6 +72,21 @@ kups_mcmc_rigid mcmc_rigid.yaml
...
+| Command | Force Field | Description |
+|---------|-------------|-------------|
+| `kups_mcmc_widom` | Lennard-Jones + Ewald | Widom ghost insertion for $\mu^\mathrm{ex}$, $K_H$, $q_\mathrm{st}$ |
+
+```sh
+cd examples
+kups_mcmc_widom mcmc_widom.yaml
+```
+
+[`finalize_widom`][kups.mcmc.widom.finalize_widom] applied to the accumulated [`WidomStatistics`][kups.mcmc.widom.WidomStatistics] yields $\mu^\mathrm{ex}$, $K_H$, and $q_\mathrm{st}$ (Vlugt 2008 eq. 16, $N=0$). Multiple hosts in one config become parallel batched systems on the `Table[SystemId, ...]` axis; duplicate a host entry $k$ times for $k$ independent chains. Blocking spheres ([`HostConfig.blocking_spheres`][kups.application.mcmc.data.HostConfig.blocking_spheres]) exclude regions from insertion.
```
n-gao: this description sound inaccurate. finalize_widom is never used. The accumulated statistics are now on disk. Also the details for batching etc and blocking spheres is not necessary here.

n-gao (changes requested):
I think there is some legacy code from the previous approach spread thoughout the implementation?

src/kups/mcmc/widom.py
```diff
@@ -0,0 +1,162 @@
...
+statistics.
+
+Contents:
+
+- [widom_test][kups.mcmc.widom.widom_test]: per-system $\ln\alpha$ for a ghost move
+- [GhostProbe][kups.mcmc.widom.GhostProbe]: propagator wrapper accumulating
+  the ratio via a lens + update callback
+- [WidomStatistics][kups.mcmc.widom.WidomStatistics]: running-sum accumulator
+  consumed by the post-hoc analyzer
+  ([analyze_widom_file][kups.application.mcmc.analysis.analyze_widom_file]).
```
n-gao: this is not in this file

src/kups/mcmc/moves.py
```diff
@@ -813,9 +814,10 @@ def _propose_insertion(
         )
         return changes, self._zero_ratio(state)
 
-    def _propose_deletion(
+    def propose_deletion(
         self, key: Array, state: State, /
     ) -> tuple[ExchangeChanges, LogProbabilityRatio]:
+        """Propose deleting a random motif. Move log-ratio is zero."""
```
n-gao: very sneakily small docstrings ;-) I prefer if they explain args and return types but fine with me ^^

n-gao (approved):
Two minor changes, I will already accept here.

PR #127, opened 2026-05-19
fix(neighborlist): undercount of periodic images in cell-list

felixmusil:
## Summary
- Fix an off-by-some bug in `_get_candidate_images` (src/kups/core/neighborlist.py) where `ceil(2 * cutoff / perp_length)` would round down the image-replication count for cutoffs that slightly exceed `perp / k`. Replaced with `2 * ceil(cutoff / perp_length) + 1`, which is always odd and large enough to fully enclose the cutoff sphere.
- Add `test/core/test_neighborlist_ase.py` that cross-checks `CellListNeighborList` and `DenseNearestNeighborList` against `ase.neighborlist.neighbor_list` across cubic, hexagonal, triclinic, and a 500-atom Cu supercell, with cutoffs spanning below, near, and well past the perpendicular cell length. The test uses `UniversalNeighborlistParameters.estimate` + `LensCapacity` so capacities self-grow via `fix_or_raise`.

The new test reproduced the bug on an 8-atom triclinic case: `Dense` and ASE agreed (64 directed edges), but `CellList` returned 56 — exactly the kind of silent undercount the formula change addresses.

## Test plan
- [x] `uv run pytest test/core/test_neighborlist_ase.py` — all 10 cases pass (Dense + CellList × 5 system/cutoff combinations).
- [x] `uv run pytest test/core/test_neighborlist.py` — existing 90 tests still green.
- [x] CI runs the full suite.

### DIFF
--- a/src/kups/application/relaxation/data.py
+++ b/src/kups/application/relaxation/data.py
@@ -6,6 +6,7 @@
 from __future__ import annotations
 
 from pathlib import Path
+from typing import Any, TypeGuard
 
 import ase
 import jax.numpy as jnp
@@ -17,7 +18,7 @@ from kups.application.utils.particles import (
     default_exclusion,
     particles_from_ase,
 )
-from kups.core.cell import Cell
+from kups.core.cell import Cell, Periodic3D
 from kups.core.data import Table
 from kups.core.data.index import Index
 from kups.core.typing import ExclusionId, ParticleId, SystemId
@@ -85,6 +86,10 @@ class RelaxParameters(BaseModel):
     """List of Optax transform specifications passed to `make_optimizer`."""
     optimize_cell: bool
     """Whether to also relax lattice vectors."""
+
+
+def _assert_periodic(x: Cell[Any]) -> TypeGuard[Cell[Periodic3D]]:
+    return x.periodic == (True,) * 3
 
 
 def relax_state_from_ase(
--- a/src/kups/application/relaxation/data.py
+++ b/src/kups/application/relaxation/data.py
@@ -6,7 +6,6 @@
 from __future__ import annotations
 
 from pathlib import Path
-from typing import Any, TypeGuard
 
 import ase
 import jax.numpy as jnp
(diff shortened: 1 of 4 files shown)


test/core/test_neighborlist_ase.py
```diff
@@ -0,0 +1,233 @@
...
+from kups.core.utils.jax import dataclass
+
+
+@pytest.fixture(autouse=True, scope="module")
+def _clear_cache():
+    jax.clear_caches()
+    gc.collect()
+    yield
+    jax.clear_caches()
+    gc.collect()
```
n-gao: we can import clear_cache like in other files

src/kups/application/relaxation/data.py
n-gao, line 1: I guess these changes slipped in accidentally?

test/core/test_neighborlist_ase.py
n-gao, line 1: These tests are fantastic and make a lot of sense. However, they are also the slowest test in our test suite. Can we reduce system sizes (+cutoffs) to make these more efficient?

Interestingly, I ran the file on main and also a Dense neighborlist test failes. Any idea why?
```
FAILED test/core/test_neighborlist_ase.py::TestNeighborListAgainstASE::test_matches_ase[DenseNearestNeighborList-cubic_Al] - AssertionError: cubic_Al @ cutoff=6.0 DenseNearestNeighborList: kups produced 216 directed edges, ase produced 234 — dis..
```

n-gao (changes requested):
The changes make sense to me, could we add an isolated test that covers exactly that behavior?

test/core/test_neighborlist_ase.py
felixmusil: the bugfix affects both the cell list and dense implementations

felixmusil: on the slowness side I don't have the time to address this in this PR. Changing the structures and cutoff is likely to reduce coverage. I can  switch to comparing with vesin (new dep that's faster) and get better guesses to the NL capacities?

n-gao (approved):
LGTM

test/core/test_neighborlist_ase.py
n-gao: I will make a followup PR that reduces the system size. all good

PR #170, opened 2026-06-04
feat: Add per-system Armijo and strong-Wolfe line-search transforms

### LINKED ISSUE #162: Add linesearch for relaxation
### Problem or Use Case

Linesearch can be more optimal than doing simple gradient updates we should support common line searches (armijo, strong_wolfe)

### Proposed Solution

_No response_

### Alternatives Considered

_No response_

n-gao:
> 🤖 _Auto-generated by Claude. Review and edit as needed._

Added per-system Armijo and strong-Wolfe line-search transforms (`ScaleByBacktrackingLinesearch` and `ScaleByZoomLinesearch`) that rescale descent directions by per-system step lengths, enabling efficient batched optimization where systems at convergence don't throttle others, and paired naturally with L-BFGS.

Closes #162

### DIFF
--- a/src/kups/relaxation/propagator.py
+++ b/src/kups/relaxation/propagator.py
@@ -24,13 +24,16 @@ from kups.relaxation.optimizer import Optimizer, apply_updates
 
 @dataclass
 class RelaxationPropagator[State, PyTree, OptState](Propagator[State]):
-    """Unified propagator for gradient-based optimization using Optax.
+    """Unified propagator for gradient-based optimization.
 
-    Uses a Potential to compute energy and gradients. Supports both standard
-    optimizers (Adam, SGD) and line-search optimizers (L-BFGS, backtracking).
+    Uses a Potential to compute energy and gradients. Supports standard optax
+    optimizers (Adam, SGD) and the per-system line searches in
+    :mod:`kups.relaxation.transforms` (backtracking, More-Thuente).
 
-    For line-search optimizers, the potential is evaluated at trial points during
-    the line search. For standard optimizers, it's evaluated once per step.
+    Each step it passes the optimizer the gradient, the current per-system
+    energies, and a ``value_and_grad_fn`` for evaluating trial points; the line
+    searches use these to evaluate the objective along the search direction,
+    standard transforms ignore them.
 
     After computing energy and gradients, the potential's patch is applied to the
     state. This allows potentials to update internal state (e.g., neighbor lists)
@@ -43,14 +46,15 @@ class RelaxationPropagator[State, PyTree, OptState](Propagator[State]):
     Attributes:
         potential: Potential that computes energy and gradients of type PyTree
         property: Lens to get/set the property being optimized
-        opt_state: Lens to get/set the Optax optimizer state
-        optimizer: Optax gradient transformation
+        opt_state: Lens to get/set the optimizer state
+        optimizer: Gradient transformation
 
     Example:
         ```python
         import optax
+        from kups.relaxation.optimizer import chain
         from kups.relaxation.propagator import RelaxationPropagator
-        from kups.core.potential import MappedPotential
+        from kups.relaxation.transforms import ScaleByAseLbfgs, ScaleByMoreThuenteLinesearch
 
         # Standard optimizer (Adam)
         propagator = RelaxationPropagator(
@@ -60,25 +64,16 @@ class RelaxationPropagator[State, PyTree, OptState](Propagator[State]):
             optimizer=optax.adam(0.01),
         )
 
-        # Line-search optimizer (L-BFGS)
+        # L-BFGS with a per-system strong-Wolfe line search
         propagator = RelaxationPropagator(
             potential=my_potential,
             property=positions_lens,
             opt_state=lens(lambda s: s.opt_state),
-            optimizer=optax.lbfgs(),
-        )
-
-        # With gradient projection
-        mapped_potential = MappedPotential(
-            full_potential,
-            gradient_map=lambda g: g.positions,
-            hessian_map=lambda h: h,
-        )
-        propagator = RelaxationPropagator(
-            potential=mapped_potential,
-            property=positions_lens,
-            opt_state=lens(lambda s: s.opt_state),
-            optimizer=optax.lbfgs(),
+            optimizer=chain(
+                ScaleByAseLbfgs(memory_size=10),
+                optax.scale(-1.0),
+                ScaleByMoreThuenteLinesearch(),
+            ),
         )
 
         state = propagator(key, state)  # One optimization step
@@ -94,13 +89,11 @@ class RelaxationPropagator[State, PyTree, OptState](Propagator[State]):
         del key
         params = self.property.get(state)
 
-        def value_fn(p: PyTree) -> Array:
-            updated_state = self.property.set(state, p)
-            result = self.potential(updated_state)
-            return result.data.total_energies.data.sum()
+        def value_and_grad_fn(p: PyTree) -> tuple[Any, PyTree]:
+            out = self.potential(self.property.set(state, p)).data
+            return out.total_energies, out.gradients
 
         potential_out = self.potential(state)
-        value = potential_out.data.total_energies.data.sum()
         grad = potential_out.data.gradients
         # Apply the patch
         energies = potential_out.data.total_energies
@@ -110,13 +103,16 @@ class RelaxationPropagator[State, PyTree, OptState](Propagator[State]):
 
         opt_state_current = self.opt_state.get(state)
 
+        # grad, energies (the current per-system energies) and value_and_grad_fn
+        # are the per-system objective the line-search transforms read; standard
+        # transforms ignore them.
         updates, new_opt_state = self.optimizer.update(
             grad,
             opt_state_current,
             params,
-            value=value,
             grad=grad,
-            value_fn=value_fn,  # necessary for line-search optimizers
+            energies=energies,
+            value_and_grad_fn=value_and_grad_fn,
         )
 
         new_params = apply_updates(params, updates)
--- a/src/kups/relaxation/transforms/linesearch.py
+++ b/src/kups/relaxation/transforms/linesearch.py
@@ -1,0 +1,580 @@
+# Copyright 2024-2026 Cusp AI
+# SPDX-License-Identifier: Apache-2.0
+
+"""Per-system line searches that reproduce ASE's implementations.
+
+Two transforms that rescale an incoming descent direction ``d`` by a
+per-system step length ``t``. Like every other transform in this package,
+the search is taken per system: ``φ(t)``, ``φ'(t)`` and the accept /
+backtrack decision are arrays of shape ``(n_systems,)`` keyed on the
+system ids, so a system that is already satisfied never throttles the step
+of one that is not, and a batched run is bit-identical to running each
+system on its own.
+
+The two searches reproduce ASE's two line searches step-for-step (a single
+system matches ASE to floating-point tolerance):
+
+* :class:`ScaleByBacktrackingLinesearch` is :class:`ase.utils.linesearcharmijo`
+  ``LineSearchArmijo``: quadratic-interpolation backtracking enforcing the
+  Armijo sufficient-decrease condition, with the ``t ← max(t_quad, t/10)``
+  floor.
+* :class:`ScaleByMoreThuenteLinesearch` is :class:`ase.utils.linesearch` ``LineSearch``:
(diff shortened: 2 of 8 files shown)


n-gao:
* **#171** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/171?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#173](https://github.com/cusp-ai-oss/kUPS/pull/173) <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/173?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#178](https://github.com/cusp-ai-oss/kUPS/pull/178) <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/178?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#170** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/170?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/170?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#206** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#205** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,378 @@
...
+    *,
+    c1: float,
+    decrease_factor: float,
+    max_steps: int,
+    t_init: float,
+) -> Array:
+    """Per-system Armijo backtracking; returns the step ``t`` per system.
+
+    Starts every descent system at ``t_init`` and multiplies by
+    ``decrease_factor`` until ``φ(t) ≤ φ(0) + c1·t·φ'(0)`` holds, freezing the
```
felixmusil: my current tests show that this simple backtracking under performs unfortunately. I suggest to make use of the gradient computed on the trial point to do a cubic interpolation (which can fail) and fall back to the quadratic interpolation. Claude knows about this `(Nocedal & Wright, eq. 3.59)` as the reference for this twist

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,426 @@
...
+    optimizer = chain(
+        ScaleByAseLbfgs(memory_size=10),   # H⁻¹∇L
+        optax.scale(-1.0),                 # descent direction d = -H⁻¹∇L
+        ScaleByZoomLinesearch(),           # t · d
+    )
+
+Each step :class:`kups.relaxation.propagator.RelaxationPropagator` supplies
+the current per-system energies (``energies``), the raw gradient (``grad``)
+and a ``value_and_grad_fn`` that returns the per-system energies and gradient
+at a trial point. A system whose direction is not a descent direction
```
jonkhler, line 55: The docstring promises this for any direction: a non-descent direction
(`∇L·d ≥ 0`) is left unmoved (`t = 0`). `ScaleByBacktrackingLinesearch` enforces
it (`descent = dphi0 < 0`, line 183), but `ScaleByZoomLinesearch` has no such
guard. With `φ'(0) ≥ 0` the zoom can bracket-expand to `t_max` and take a large
step instead of returning 0. Repro: concave-along-the-line objective with
`φ'(0)=+0.2` returns `t ≈ t_max`, while backtracking returns 0 on the same input.
This matters if L-BFGS transiently emits a non-descent direction.

Either add a `dphi0 < 0` guard to `_zoom` (seed those systems at `t=0`, never move
them), or scope this docstring sentence to the backtracking transform.

```python
def test_zoom_non_descent_nonconvex_does_not_move():
    """Non-descent direction (φ'(0) > 0) must yield t = 0 even for non-convex φ."""
    opt = ScaleByZoomLinesearch()
    idx = _system_index([0], 1)
    x = jnp.array([0.0])
    direction = jnp.array([1.0])  # φ'(0) = grad(0)·d = +0.2 > 0

    def vag(p):
        return idx.sum_over(-(p**2) + 0.2 * p), (-2.0 * p + 0.2)

    step, _ = _update(opt, opt.init(x, index_prefix=idx), x, direction, vag)
    npt.assert_allclose(step, jnp.zeros(1), atol=1e-12)
```

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,426 @@
...
+        # Zoom: shrink the upper bound.
+        hi = jnp.where(z_shrink, t, hi)
+        # Zoom: advance the lower bound (flip hi←lo when the slope points outward).
+        flip = z_advance & (dphi * (hi - lo) >= 0)
+        hi = jnp.where(flip, lo, hi)
+        lo = jnp.where(z_advance, t, lo)
+        phi_lo = jnp.where(z_advance, phi, phi_lo)
+
+        accept = b_accept | z_accept
+        t_acc = jnp.where(accept, t, jnp.where(active & armijo, t, c["t_acc"]))
```
jonkhler: `t_acc` keeps the most recent Armijo step. N&W's bracket-low / safe point is the
sufficient-decrease step with the *smallest* value (property (b)); optax keeps
this in `safe_stepsize`, updating only when the value drops. With strict curvature
and a small `max_steps` the returned energy is non-monotonic in the budget (e.g.
budget 4 gives `φ=0.38`, budget 5 gives `φ=1.33`), so the non-converged fallback
can return a worse point than one already found. The accepted-Wolfe path is
unaffected.

Optional: overwrite `t_acc` only when the new Armijo point has lower energy.

```python
def test_zoom_fallback_tracks_best_not_last():
    """Non-converged zoom: returned energy must not rise as the budget grows."""
    hess = jnp.array([1.0, 8.0])
    x = jnp.array([1.0, 1.0])
    idx = _system_index([0, 0], 1)
    direction = -hess * x

    def energy_at(t):
        return float(jnp.sum(0.5 * hess * (x + t * direction) ** 2))

    phis = [
        energy_at(
            float(
                _single_step(
                    ScaleByZoomLinesearch(c2=1e-4, max_steps=k),
                    x, hess=hess, idx=idx, direction=direction,
                )[0] / direction[0]
            )
        )
        for k in range(3, 11)
    ]
    assert all(b <= a + 1e-9 for a, b in zip(phis, phis[1:])), phis
```

src/kups/relaxation/propagator.py
```diff
@@ -110,13 +103,16 @@ def value_fn(p: PyTree) -> Array:
...
+        # are the per-system objective the line-search transforms read; standard
+        # transforms ignore them.
         updates, new_opt_state = self.optimizer.update(
             grad,
             opt_state_current,
             params,
-            value=value,
             grad=grad,
-            value_fn=value_fn,  # necessary for line-search optimizers
+            energies=energies,
```
jonkhler, line 114: Dropping `value`/`value_fn` here (now passing `energies`/`value_and_grad_fn`)
breaks optax-native line searches reached via the `hasattr(optax, name)` fallback
in `config.get_transform` (config.py:70). The `scale_by_*_linesearch` names are
shadowed by the custom registry, but `optax.lbfgs` is not: a config
`{"transform": "lbfgs"}` resolves to `optax.lbfgs()`, which then raises at runtime
`TypeError: ... missing 2 required keyword-only arguments: 'value' and 'value_fn'`
(reproduced for `lbfgs`, `scale_by_zoom_linesearch`,
`scale_by_backtracking_linesearch`).

Either reject such transforms in `get_transform` with a clear error, or also pass
`value` + a scalar `value_fn` here to keep them working (with their tree-global,
batch-unsafe behaviour, i.e. the thing #94 fixed for the per-system path).

```python
def test_optax_native_linesearch_through_chain():
    """A config resolving to optax.lbfgs must not crash through the chain."""
    import optax
    from kups.relaxation.optimizer import chain

    idx = _system_index([0, 0], 1)
    x = jnp.array([1.0, 1.0])
    vag = _value_and_grad_fn(jnp.ones(2), idx)
    energies, grad = vag(x)
    opt = chain(optax.lbfgs())
    opt.update(grad, opt.init(x), x, grad=grad, energies=energies, value_and_grad_fn=vag)
```

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,426 @@
...
+    }
+
+    def cond(c: dict[str, Array]) -> Array:
+        return (c["i"] < max_steps) & ~jnp.all(c["done"])
+
+    def body(c: dict[str, Array]) -> dict[str, Array]:
+        t = c["t"]
+        phi, dphi = evald(t)
+        active = ~c["done"]
+        armijo = phi <= phi0 + c1 * t * dphi0
```
jonkhler: Note, no change needed. The folded test `phi <= phi0 + c1*t*dphi0` is more robust
to float32 cancellation than optax's error-vs-tol form: with a 1e7 energy baseline
in float32 this converged, while optax with its Hager-Zhang criterion disabled
stalled at `t=0`. There is no approximate-decrease criterion, but I could not
break it. If tight relaxation runs in float32, a doc note to prefer x64 is enough.

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,426 @@
...
+        accept = b_accept | z_accept
+        t_acc = jnp.where(accept, t, jnp.where(active & armijo, t, c["t_acc"]))
+        return {
+            "i": c["i"] + 1,
+            "done": c["done"] | accept,
+            "t_acc": t_acc,
+            "zooming": c["zooming"] | b_zoom1 | b_zoom2,
+            # Next trial: expand while bracketing, else the new bracket midpoint.
+            "t": jnp.where(
+                b_expand, jnp.minimum(expand_factor * t, t_max), 0.5 * (lo + hi)
```
jonkhler: Note, perf only. The zoom uses bisection (`0.5*(lo+hi)`), which is valid per N&W
and self-safeguarding, but slower to meet the curvature condition than the
cubic/quad interpolation optax uses and that `_interp_step` already implements for
backtracking. Optional: reuse `_interp_step`-style interpolation in `_zoom` for
fewer evaluations.

jonkhler (changes requested):
The per-system line searches match the strong-Wolfe bracket/zoom and the
interpolating Armijo backtracking in Nocedal & Wright (Alg 3.5/3.6, eq 3.58/3.59);
I checked every state transition against the book and against optax 0.2.8, and
cross-checked numerically (backtracking lands on the same step as optax's zoom to
6 dp; both satisfy the conditions). The per-system reductions decouple systems as
intended. A few points below: a real behaviour gap (the zoom and the docstring
disagree on non-descent directions), a non-converged-path fallback choice, a
runtime break for optax-native line searches reached via config, plus two notes.

src/kups/relaxation/transforms/linesearch.py
```diff
@@ -0,0 +1,426 @@
...
+            ``0 < min_shrink < max_shrink < 1``.
+        max_steps: Maximum backtracking iterations.
+        t_init: Initial trial step. ``1.0`` suits Newton-scaled directions.
+    """
+
+    c1: float = field(static=True, default=1e-4)
+    min_shrink: float = field(static=True, default=0.1)
+    max_shrink: float = field(static=True, default=0.5)
+    max_steps: int = field(static=True, default=20)
+    t_init: float = field(static=True, default=1.0)
```
felixmusil, line 475: in [`ase/utils/linesearcharmijo.py:431-437`](vscode-webview://0ujg0rulqc5pq7mefam3dpvqlus6qedomvjuch14g4elf5r7apl9/third-party/resolves/python-default/.pixi/envs/default/lib/python3.12/site-packages/ase/utils/linesearcharmijo.py#L431-L437)  (so specific to armijo) they get an estimate for `t_init` and use it if it is <0.5 or >1.5 (the rounding happens `linesearcharmijo.py:L280-L281`). I found it quite important to reduce the number of evals in the linesearch and the overall number of steps when using the Exp preconditioner.

n-gao:
I updated the code and tests to match ASE's line search 1:1. I also added code to reject optax's line searches since those would not behave as intended.

jonkhler (approved):
lgtm

n-gao:
### Merge activity

* **Jun 24, 12:44 PM UTC**: @n-gao merged this pull request with [Graphite](https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/170).

PR #171, opened 2026-06-04
refactor: Rename `lh`/`rh`/`for_indices` to `keys`/`queries`/`queried_keys` in neighbor list API

n-gao:
> 🤖 *Auto-generated by Claude. Review and edit as needed.*

# PR Description

Renamed neighbor list API parameters from `lh`/`rh`/`for_indices` to `keys`/`queries`/`queried_keys` respectively to improve semantic clarity: `keys` represents the self-graph/output table, `queries` denotes bipartite query tables, and `queried_keys` specifies affected particle indices for self-graph updates.

PR #173, opened 2026-06-04
feat: Add `Lens.then` for post-write state invariants via `ThenLens`

### LINKED ISSUE #172: Proposal: Endo[S] + Lens.then for write-side state invariants
## Summary

Add a write-side combinator to the lens library so a lens can re-establish a whole-state invariant after every write, without affecting reads:

- a type alias `Endo[S] = View[S, S]` (an endomorphism on the state), and
- a `then` combinator on `Lens`: `set(s, r) = after(inner.set(s, r))`, with `get` unchanged.

## Motivation

Variable-cell relaxation optimizes atomic positions and the cell together. When the optimizer writes a new cell, the atoms should ride with it — i.e. fractional coordinates are held fixed and Cartesian positions are recomputed as `positions = cell.to_real(fractional)`. Otherwise a box deformation throws every atom off its lattice site, which couples the position and cell subproblems badly.

The relaxation `property` is a plain lens onto `(positions, cell)`, so setting the cell leaves the Cartesian positions untouched and the invariant `positions == cell.to_real(fractional)` is silently violated. We want a lens whose *write* restores that invariant while its *read* stays a plain projection.

More generally: state updates often have derived/coupled fields that must be refreshed after a focused write. A write-side hook expresses this directly instead of scattering fix-ups across call sites or overloading the getter.

## Proposed API

```python
type Endo[S] = View[S, S]
"""An endomorphism on the state, post-composed with a lens's writes."""
```

On the `Lens` protocol (default-implemented in `BaseLens`, alongside `focus` / `merge` / `nest` / `at`):

```python
def then[S2](self: Lens[S2, R], after: Endo[S2]) -> Lens[S2, R]:
    """Re-establish a whole-state invariant after every write.

    Returns a lens with the same getter, whose setter post-composes the
    write with ``after``: ``set(s, r) == after(inner.set(s, r))``. The
    normalization is one-directional — reads are unaffected.
    """
```

`apply`, `at`, `merge`, and `nest`-derived writes inherit the behavior for free, since `BaseLens` routes them through `set`.

## Semantics and contract

`after` is an `Endo[S]` applied to the **post-set** state only — it never sees the pre-set state. So it must:

- depend only on the post-set state (store the independent DOF, e.g. fractional coords + cell, and derive the dependent one, e.g. Cartesian positions);
- be idempotent;
- write only fields disjoint from this lens's focus.

Under those conditions the lens laws hold on the consistent submanifold:

- **PutGet** holds because `after` touches only out-of-focus fields.
- **GetPut** holds whenever the state is already consistent (`after` is an idempotent fixpoint there).

Endomorphisms form a monoid under composition, so multiple invariants combine by composing their `Endo[S]` into one before passing to `then`.

## Example

```python
def add_lattice_dependence(s: State) -> State:
    cell = s.systems[s.particles.data.system].cell
    frac_pos = cell.to_fractional(s.particles.data.positions)
    real_pos = cell.to_real(jax.lax.stop_gradient(frac_pos))
    new_pos = x.particles.data.positions + (real_pos - jax.lax.stop_gradient(real_pos))
    return bind(s, lambda x: x.particles.data.positions).set(new_pos)

property = prop_lens.then(add_lattice_dependence)
```

## Scope / non-goals

`then` fixes the **primal** write: the state update and any energy re-evaluation through `property.set(...)` become consistent. The **gradient** handed to the optimizer arrives in Cartesian/partial coordinates `(∂E/∂r, ∂E/∂h|_r)` and must independently be expressed in the same `(fractional, cell)` coordinates (the matching cotangent: `∂E/∂s = Gʳ @ hᵀ`, `dE/dh|_s = ∂E/∂h|_r + h⁻ᵀ·Σ r⊗Gʳ`). That cotangent wiring is tracked separately; this issue covers only the lens combinator.

## Alternatives considered

- A pre/post hook `(S_old, S_new) -> S_new`: strictly more general, but heavier, and unnecessary once the independent DOF is stored and the dependent one derived.
- Domain-specific names (`normalize`, `reconcile`, `rectify`): rejected in favor of the generic `Endo` / `then`, which name the type and the composition rather than editorializing about intent.

n-gao:
Adds a `then` combinator to `Lens` and `BoundLens` that post-composes a state endomorphism (`Endo[S]`) with every write, allowing whole-state invariants to be re-established after each `set`, `apply`, `at`, or `merge` call. Reads are unaffected.

The implementation introduces:

- The `Endo[S]` type alias for a `View[S, S]` (a function from state to state).
- `ThenLens`, a new `BaseLens` subclass that delegates `get` to its inner lens and wraps `set` with the provided endomorphism.
- `then` on `BaseLens` constructing a `ThenLens`, and on `SimpleBoundLens` delegating to the unbound lens before rebinding.
- `then` on both `Lens` and `BoundLens` protocols.

Multiple `then` calls chain in left-to-right order, so dependent invariants (e.g. recomputing `cartesian` before recomputing `total`) compose correctly. Tests cover single and chained invariants, the one-directional read guarantee, `apply`, bound lens usage, and JIT compatibility.  
  
Closes #172

### DIFF
--- a/src/kups/core/lens.py
+++ b/src/kups/core/lens.py
@@ -116,8 +116,14 @@ class Update[S, R](Protocol):
     def __call__[S2](self: Update[S2, R], state: S2, /, value: R) -> S2: ...
 
 
-type Modifier[R] = Callable[[R], R]
-"""Type alias for a function that transforms a value."""
+type Endo[X] = View[X, X]
+"""An endomorphism: a function from a type back into itself.
+
+Used as the focused-value transform in [`apply`][kups.core.lens.Lens.apply] and
+as the post-set state transform in [`then`][kups.core.lens.Lens.then].
+Endomorphisms form a monoid under composition, so several can be combined into
+one before being passed in.
+"""
 
 
 class _NOT_SET_TYPE(enum.Enum):
@@ -209,7 +215,7 @@ class Lens(Protocol[S, R]):
         """
         ...
 
-    def apply[S2, R2](self: Lens[S2, R2], state: S2, /, modifier: Modifier[R2]) -> S2:
+    def apply[S2, R2](self: Lens[S2, R2], state: S2, /, modifier: Endo[R2]) -> S2:
         """Apply a modifier function to the focused value.
 
         Args:
@@ -270,6 +276,26 @@ class Lens(Protocol[S, R]):
 
         Returns:
             A new lens that composes this lens with the provided lens/view
+        """
+        ...
+
+    def then[S2](self: Lens[S2, R], after: Endo[S2]) -> Lens[S2, R]:
+        """Re-establish a whole-state invariant after every write.
+
+        Returns a lens with the same getter whose setter post-composes the
+        write with ``after``: ``set(s, r) == after(set(s, r))``. Reads are
+        unaffected — the rewrite is one-directional. ``apply``/``at``/``merge``
+        inherit it because they route through ``set``.
+
+        ``after`` only sees the post-set state, so it must depend on that state
+        alone, be idempotent, and write only fields disjoint from this lens's
+        focus (so reads still round-trip).
+
+        Args:
+            after: State endomorphism applied after each write.
+
+        Returns:
+            A lens whose writes restore the invariant.
         """
         ...
 
@@ -345,7 +371,7 @@ class BoundLens(Protocol[S, R]):
         """
         ...
 
-    def apply[R2](self: BoundLens[S, R2], modifier: Modifier[R2]) -> S:
+    def apply[R2](self: BoundLens[S, R2], modifier: Endo[R2]) -> S:
         """Apply a modifier function to the focused value in the bound data structure.
 
         Args:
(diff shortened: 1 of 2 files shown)


n-gao:
* **#173** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/173?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/173?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#171** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/171?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#178](https://github.com/cusp-ai-oss/kUPS/pull/178) <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/178?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#170** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/170?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#206** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/206?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#205** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/205?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #179, opened 2026-06-08
fix: preserve inner lens setter in Lens.nest

### LINKED ISSUE #174: Lens.nest / focus discard the inner lens's custom setter
`Lens.nest(other)` accepts a full `Lens` but uses only `other.get`, then re-infers the setter by path-tracing. `other.set` is silently discarded.

`src/kups/core/lens.py:425`
```python
def nest(self, other):
    view_func = other.get if isinstance(other, Lens) else other
    return self.focus(view_func)   # -> NestedLens(self, SimpleLens(view(other.get)))
```

## Repro (fails today)
```python
inner = LambdaLens(                                  # set energy AND zero positions
    lambda s: s.energy,
    lambda s, value: replace(s, energy=value, positions=s.positions * 0),
)
outer = lens(lambda o: o.inner, cls=Outer)

outer.nest(inner).set(state, 9.0).inner.positions        # UNCHANGED  (setter dropped)
NestedLens(outer, inner).set(state, 9.0).inner.positions # zeroed     (correct)
```
The `NestedLens` constructor threads both setters; only the `nest`/`focus` methods fail to route to it.

`IndexLens` is worse than silent: `nest`ing it raises, because its getter (`arr.at[idxs].get()`) is a computed value that path-tracing rejects.

## Fix
```python
def nest[U, R2](self: Lens[S, R2], other: Lens[R2, U] | View[R2, U]) -> Lens[S, U]:
    if isinstance(other, Lens):
        return NestedLens(self, other)   # preserve both halves
    return self.focus(other)             # bare view -> infer the setter
```
Widen the protocol signatures (`Lens.nest`, `BoundLens.nest`) to `Lens | View`. `focus` stays getter-only by design. `SimpleBoundLens.nest` delegates, so it inherits the fix.

## Blast radius
`.nest(` is used once in-tree (`SimpleBoundLens.nest`, internal); no caller composes a custom-setter lens. For `SimpleLens` (the only kind composed today) `NestedLens(self, other).set` equals the inferred set, so zero behavior change. The repro flips red→green; promote it to `test/core/test_lens.py` alongside a `nest(IndexLens)` case (raises today).

cc @n-gao

jonkhler:
`nest` kept only the inner lens's getter and re-inferred a setter via `focus` path-tracing, dropping custom setters; `IndexLens`'s computed getter made it raise.

Now returns `NestedLens(self, other)` (threads get and set) and rejects non-`Lens` args with `TypeError`. `SimpleBoundLens.nest` inherits via delegation.

Closes #174

### DIFF
--- a/src/kups/core/lens.py
+++ b/src/kups/core/lens.py
@@ -257,17 +257,17 @@ class Lens(Protocol[S, R]):
         ...
 
     def nest[U, S2, R2](self: Lens[S2, R2], other: Lens[R2, U]) -> Lens[S2, U]:
-        """Nest another lens or view within this lens.
+        """Compose this lens with an inner lens, preserving the inner lens's setter.
 
-        This provides an alternative to focus() that works with both lenses and views.
-        When given a lens, it extracts the view from it; when given a view directly,
-        it uses it as-is.
+        Unlike focus(), which infers a setter by tracing the path of a getter, nest
+        threads the inner lens's own get and set. For a bare getter (no setter),
+        use focus() instead.
 
         Args:
-            other: Either a lens or view to nest within this lens
+            other: The inner lens to compose within this lens
 
         Returns:
-            A new lens that composes this lens with the provided lens/view
+            A new lens that composes this lens with the inner lens
         """
         ...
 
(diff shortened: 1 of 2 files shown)


src/kups/core/lens.py
```diff
@@ -423,8 +423,12 @@ def merge[S2, R2](
         return MergedLens(self, other)
 
     def nest[U, R2](self: Lens[S, R2], other: Lens[R2, U]) -> Lens[S, U]:
-        view_func = other.get if isinstance(other, Lens) else other
-        return self.focus(view_func)
+        if not isinstance(other, Lens):
+            raise TypeError(
+                f"nest requires a Lens, got {type(other).__name__}. "
+                "Use .focus(view) to compose a bare getter."
+            )
```
n-gao, line 430: isinstance checks are not very stable for protocols. (we also do it nowhere else where the type hint is explicit?)

n-gao (approved):
LGTM  
  
Small question regarding the isinstance check that we do nowhere else :)

Issue #190, opened 2026-06-11
Additional D2H overhead for small systems can be removed.


snme:
### Problem or Use Case

I noticed that `propagate_and_fix` forces a host->device->host round trip at each invocation, and thus at every step of `run_simulation_cycles`, which effectively serializes the host and the gpu. This is caused by 
1. a blocking D2H sync on line 169 of propagator.py: checking `.failed_assertions` , 
2. which calls `all_assertions_pass.item()`, 
3. which blocks until the GPU finishes the step, 
4. which blocks the next step from getting dispatched until step N completes, 
5. which inhibits the pipelining from being async). 
I originally noticed this when simulating a small number of atoms (~1000) with a LJ + Ewald potential, performance monitoring with py-spy (Fig. 1, below): 
<img width="518" height="538" alt="Fig. 1" src="https://github.com/user-attachments/assets/26e3a4b8-ef09-490e-9c45-80d0f85bdebc" />
You can see that a huge amount of time for small systems is being taken up by this D2H block in `propagate_and_fix`.

I measured this delay to be a fixed cost (on an L4 node with a LJ + Ewald potential), invariant to system size, of **1.7-1.9ms per step**.

I acknowledge that we incur this overhead because we want the propagator to catch incorrect states as soon as possible. However, as measured by the plots below, this overhead makes the biggest difference (**up to ~20% slowdowns**) for _small_ system sizes. At these small system sizes, the cost of throwing away a few erroneous steps is nearly negligible, and at large system sizes, the cost + benefit of this overhead is also nearly negligible. Therefore, I argue that it's beneficial to remove this overhead.
<img width="910" height="364" alt="Image" src="https://github.com/user-attachments/assets/4ff6292a-80d2-40d6-bbfa-b89610c141f3" />

### Proposed Solution

I propose to swap the propagator to a LoopPropagator that checks the state assertions every _N_ steps. As shown above, doing so with a small _N_ produces an greater than 16% speedup for small systems, and makes a negligible difference for larger systems. It's a <5 line fix, and **mirrors the exact convention set by MCMC's propagator loop.**

### Alternatives Considered

See acknowledgements above.

PR #191, opened 2026-06-11
feat: Extend blocked iteration to MD

snme:
Problem: Issue #190
Feature: Just like MCMC, allow classical MD to run several steps at a time before checking for error assertions. This is called a "blocked" mode of iteration in this PR. `run_simulation_cycles` and `propagate_and_fix` are updated by building a looppropagator + overloading return types of dependent functions. The only assumption this introduces for the user is regarding the new `block_size` parameter in `MdRunConfig`, which configures how many steps to do at once before assertions are checked: the total number of steps must be divisible by the `block_size`. 
Why should we merge this? This fix enables a ~40% speedup for running molecular dynamics on small systems. This is comparable to the speedup you receive if you log 20x less frequently. For larger numbers of atoms, this does not have an effect. This is because checking assertions is a costly D2H sync with a fixed cost; checking less often ammortizes a single expensive check over several steps. In practice, this doesn't make a difference for large systems, but for small systems, in which A) the fixed cost is a larger percentage of the total simulation time, and B) throwing away a few frames and re-trying is less expensive overall, an improvement in simulation wallclock times would be impactful.
blocked_md_framing_extra_frequencies

### DIFF
--- a/src/kups/application/utils/propagate.py
+++ b/src/kups/application/utils/propagate.py
@@ -51,7 +51,9 @@ def make_cycle_function[State](
     Returns:
         A jitted ``(key, state) -> Result`` cycle function.
     """
-    return jit(as_result_function(LoopPropagator(propagator, block_size)), donate_argnums=(1,))
+    return jit(
+        as_result_function(LoopPropagator(propagator, block_size)), donate_argnums=(1,)
+    )
 
 
 def run_warmup_cycles[State](
--- a/src/kups/application/utils/propagate.py
+++ b/src/kups/application/utils/propagate.py
@@ -8,13 +8,15 @@ MD, MCMC, and relaxation application modules.
 """
 
 import logging
-from typing import Callable, Protocol
+from operator import itemgetter
+from typing import Any, Callable, Protocol
 
 import tqdm
 from jax import Array
 
 from kups.core.logging import Logger
 from kups.core.propagator import (
+    LoopPropagator,
     Propagator,
     propagate_and_fix,
     propagator_with_assertions,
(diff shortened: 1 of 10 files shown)


src/kups/application/utils/propagate.py
```diff
@@ -71,33 +92,56 @@ def run_warmup_cycles[State](
...
-            logger.log(state, i)
+        for start in range(0, num_cycles, block_size):
+            if blocked:
+                state, frames = propagate_and_fix(
+                    cycle_fn, next(chain), state, state_of=itemgetter(0)
+                )
+                logger.log_block(frames, start)
+            else:
+                state = propagate_and_fix(cycle_fn, next(chain), state)
+                logger.log(state, start)
```
snme: we are now removing the functionality of running a chunk (block) of steps at a time while also being able to log every step. instead, we will output 1 frame per chunk to mirror the output signature/maintain the same api as the non-looped-propagator. this is an appropriate trade off of slightly more functionality for way better code complexity + better maintainability of the core of MD simulations.

src/kups/application/utils/propagate.py
```diff
@@ -33,20 +34,26 @@ class CycleFunction[State](Protocol):
...
         propagator: Step propagator to compile.
+        block_size: Steps fused per dispatch; ``1`` checks assertions every step.
 
     Returns:
         A jitted ``(key, state) -> Result`` cycle function.
     """
-    return jit(as_result_function(propagator), donate_argnums=(1,))
+    return jit(
+        as_result_function(LoopPropagator(propagator, block_size)), donate_argnums=(1,)
+    )
```
n-gao: Can we move the LoopPropagator into the standard propagator initialization? That should be a smaller edit?

n-gao (changes requested):

jonkhler (approved):
conditional on @n-gao's change request lgtm

as discussed during lunch keeping the interface lean and only storing the last frame is fine

snme:
@n-gao's changes have been implemented! cc @jonkhler

n-gao (approved):
LGTM

snme:
@n-gao re: merging, should i wait until the 06-17 branch is merged into main (would avoid other pr's needing to rebase), or merge now? tysm!

PR #193, opened 2026-06-15
feat: add csvr_npt_1eval integrator (drop redundant post-barostat eval)

jonkhler:
## What

`make_csvr_npt_step` gains a `refresh_after_barostat` flag (default `True`, behaviour unchanged). With it `False`, the post-barostat force/stress evaluation is dropped and the next step's first half-kick reuses the pre-rescale forces, so each step costs **one derivative evaluation instead of two**. Pressure control is unaffected either way: the barostat consumes the fresh pre-rescale stress.

Exposed to configs as the `csvr_npt_1eval` integrator key, which shares `csvr_npt`'s parameters and state protocol (implemented by widening the existing `csvr_npt` overloads rather than adding two more to an already 9-overload function).

## Accuracy caveat

The force error grows with `|mu - 1|`, so this is only accurate for slow barostats and isotropic cells. This is documented on the flag.

## Tests

- The duplicated "1eval" physics tests are folded into the existing CSVR-NPT tests via `parametrize`.
- The old `len(one.propagators) == len(two.propagators) - 1` structural check is replaced with a test of the actual claim: 2 derivative evaluations per step → 1.

## Notes

- Rebased onto current `main` (post-#211 unified `md.py`, integrators in `kups/application/md/integrators.py`); green on ruff, pyrefly, pre-commit, and `test/application/ test/md/` (202 passed).

### DIFF
--- a/src/kups/application/md/data.py
+++ b/src/kups/application/md/data.py
@@ -261,7 +261,12 @@ class MdParameters(BaseModel):
     minimum_scale_factor: float
     """Minimum allowed box scaling factor per barostat step (dimensionless)."""
     integrator: Integrator
-    """Integration algorithm to use."""
+    """Integration algorithm to use.
+
+    ``"csvr_npt_1eval"`` is ``"csvr_npt"`` with the post-barostat force/stress eval
+    dropped, halving the derivative evaluations per step; it is only accurate for slow
+    barostats and isotropic cells. See `make_csvr_npt_step`.
+    """
     initialize_momenta: bool = False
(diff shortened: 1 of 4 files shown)


jonkhler:
* **#193** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/193?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/193?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #194, opened 2026-06-15
feat: wire the Verlet-skin neighbor list into MD

jonkhler:
**Stack (bottom → top):** #261 → #262 → #263 → #264 → #194

Top of the **Verlet-skin stack** (assert-repair tests → neighborlist_factory → margin bound → skin machinery → **this PR**).

## What

`md.verlet_skin > 0` builds the potential with `neighborlist_factory=skin_neighborlist`, seeds the Verlet fields on `MdState` via `seed_verlet_state`, and runs one ordinary propagator (`make_verlet_md_propagator`: rebuild-if-flagged → MD step → trigger/backstop) through the plain `run_md` driver at **any `block_size`** — no host-side dispatch, no per-step device→host sync, no special run loop. The rebuild/trigger steps don't consume the PRNG chain, so the noise matches a `verlet_skin = 0` run key for key.

## Tests

- Trajectory equivalence vs the every-step dense path with identical keys, per-step **and** fused (`block_size=5`), for `csvr` and `csvr_npt`.
- Mid-run capacity repair from an undersized seed (both sabotage variants: `avg_edges`-only, which makes the eager seed overgrow, and all-three).
- The escalation path (`verlet_skin` too small to absorb one step → `ValueError` with a configuration hint).
- Entry point end to end at `block_size` 1 and 3 (seeding, factory injection, HDF5 readable by the analyzer).

## Notes for the stack

The design rationale, the completeness bound, and the review history (an adversarial multi-agent review confirmed and led to fixes for cross-system margin coupling, an eager-seed shape desync, and corrected revert-semantics documentation; a fuzzer compared the refined list against brute-force image enumeration over 300+ boundary-adversarial trials with zero soundness violations) live in the lower PRs of this stack.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #195, opened 2026-06-15
feat: add fused real-space nonbonded (LJ + short-range Ewald) potential

jonkhler:
Combines the two real-space potentials kups builds separately — `make_lennard_jones_from_state` and the short-range term of `make_ewald_from_state` — into one `PotentialFromEnergy` that builds a single radius graph (one neighbor-list query, one edge gather) and runs one `jax.vjp` over the summed energy. Forces and NPT stress come from autodiff through the shared gradient lens.

Two energy bodies: `nonbonded_energy` (reuses the existing energy fns on the shared graph; numerics identical to the two summed) and `fused_nonbonded_edge_energy` (single edge gather/r², both terms inline). Full-recompute only.

Tests: `test/potential/test_nonbonded.py` (3 pass; energy/forces/stress match the separate-component oracle, both bodies).

Stacked on #194.

### DIFF
--- a/src/kups/potential/classical/__init__.py
+++ b/src/kups/potential/classical/__init__.py
@@ -50,6 +50,12 @@ from .lennard_jones import (
     make_pair_tail_corrected_lennard_jones_potential,
 )
 from .morse import MorseBondParameters, make_morse_bond_potential
+from .nonbonded import (
+    NonbondedParameters,
+    fused_nonbonded_edge_energy,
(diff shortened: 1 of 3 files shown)


jonkhler:
Closing as stale: the 192→197 PR stack was dissolved — #193, #194 and #196 were rebased onto main as independent branches, so this PR's base history has been rewritten. Reopen against main if the fused real-space nonbonded path is still wanted.

PR #196, opened 2026-06-15
feat: select PME for the Ewald long-range term via PMESettings

jonkhler:
**Stack (bottom → top):** #265 → #196

Top of the **PME stack** (smooth-PME module → **this PR**).

## What

`make_ewald_potential` / `make_ewald_from_state` accept `pme=PMESettings(mesh=..., order=...)`; omitting it leaves the direct-Ewald path **byte-for-byte unchanged**. `PMESettings` collapses what were four loose `pme_*` kwargs, making `mesh` required by construction (it lives in `pme.py`; `ewald.py` imports it under `TYPE_CHECKING` only, since pme imports ewald at runtime).

## Known limitation (documented, not blocking)

PME cannot provide the O(1) per-move structure-factor updates the MC paths rely on, so combining `pme` with a `cache_lens` asserts instead of silently turning every trial move into a full mesh evaluation. No caller enables PME on the MC paths today; the interface-level fix is left open.

## Validation

`test/potential/` green; MC smoke (`test_mcmc_rigid`, `test_mcmc_widom`) green — the default path is untouched.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #197, opened 2026-06-15
feat: add forces-only NVE/NVT path (skip dE/dcell)

jonkhler:
Adds an opt-in `forces_only` mode that differentiates the energy w.r.t. positions only, skipping the cell-virial (`dE/dcell`). For NVE/NVT, where the barostat stress is never used, this avoids the autodiff cell-gradient entirely.

Threaded through `make_nonbonded_from_state` / `make_ewald_from_state` (positions-only gradient lens) and `make_md_propagator` / `make_verlet_md_propagators` (the step neither maps nor caches the cell-virial). Default is off, so existing behavior is unchanged; raises if combined with an NPT integrator. When on, `cell_gradients` is not refreshed, so logged stress/pressure is unavailable.

The forces are bit-identical to the both-gradients path (`dE/dr` does not depend on whether `dE/dcell` is also computed). Full suite green at the stack tip (39 tests).

Stacked on #196.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,11 @@ docs/reference
 .vscode/
 *.pt
 *.model
+
+# local development
+.bedrock/
+plans/
+bench/
+benchmarks/
+.vscode-annotator/
+.mcp.json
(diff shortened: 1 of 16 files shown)


jonkhler:
Closing as stale: the 192→197 PR stack was dissolved — #193, #194 and #196 were rebased onto main as independent branches, so this PR's base history has been rewritten. Reopen against main if the forces-only path is still wanted.

PR #220, opened 2026-06-25
feat: Extend blocked iteration to MD

snme:
Problem: Issue #190
Feature: Just like MCMC, allow classical MD to run several steps at a time before checking for error assertions. This is called a "blocked" mode of iteration in this PR. `run_simulation_cycles` and `propagate_and_fix` are updated by building a looppropagator + overloading return types of dependent functions. The only assumption this introduces for the user is regarding the new `block_size` parameter in `MdRunConfig`, which configures how many steps to do at once before assertions are checked: the total number of steps must be divisible by the `block_size`. 
Why should we merge this? This fix enables a ~40% speedup for running molecular dynamics on small systems. This is comparable to the speedup you receive if you log 20x less frequently. For larger numbers of atoms, this does not have an effect. This is because checking assertions is a costly D2H sync with a fixed cost; checking less often ammortizes a single expensive check over several steps. In practice, this doesn't make a difference for large systems, but for small systems, in which A) the fixed cost is a larger percentage of the total simulation time, and B) throwing away a few frames and re-trying is less expensive overall, an improvement in simulation wallclock times would be impactful.
blocked_md_framing_extra_frequencies

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ dependencies = [
     "msgpack>=1.1.2",
     "nanoargs>=0.1.1",
     "slub>=0.1.1",
+    "matplotlib>=3.10.5",   # python 3.14 support - ASE depends on matplotlib
 ]
 
 [project.optional-dependencies]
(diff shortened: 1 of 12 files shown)


n-gao (approved):
LGTM

Issue #176, opened 2026-06-04
Optics clarity: naming + docstring cleanups (proposal)


jonkhler:
Independent clarity items in `src/kups/core/lens.py`. Each is a standalone commit.

Out of scope: `SimpleLens` rename (constructed in ~15 `src/` sites), lens variance (separate issue), the dual `__call__` get/set (load-bearing: lets a lens be passed where a `View` or `Update` is expected).

## Terminology
- **`apply` → `modify`.** The "read focus, apply fn, write back" op is `over` (Haskell) / `modify` (python-lenses). `apply` is ambiguous (reads like "invoke the lens"). ~17 `src/` call sites + 4 method defs. `modify` matches python-lenses, which the module already aligns with (`bind`, `flip`).
  ```python
  lens.apply(state, fn)   ->   lens.modify(state, fn)
  ```

## Naming
- **`LambdaLens` -> `ExplicitLens`.** Not about lambdas; it is the lens with an explicit getter+setter (callers pass named functions: `LambdaLens(view(get_name), set_name)`). Not imported outside `lens.py` + `test_lens.py`.

## Docstrings that lie
- **`all_where_lens(..., target_cls=...)`** docstring says *"(currently unused)"* (`:701`) and pyright flags it not-accessed (`:694`). It is not dead; it binds the `Target` typevar for inference (like `cls` on `lens()`); `all_isinstance_lens` passes it to get `Lens[S, tuple[cls, ...]]`. Fix: correct the docstring to "type-inference only, discarded at runtime" and add `_ = target_cls` (the pattern `lens()` uses at `:766`).
- **`conditional` -> `predicate`** in `all_where_lens`: it is a boolean predicate, and `where` already means *getter* elsewhere in the module.

## Trivia (skip unless already in those lines)
- `_NOT_SET_TYPE.OBJ`: member name `OBJ` is meaningless; `NOT_SET` reads better.
- `bind()` no-`where` branch uses `lambda x: x` where `identity` is already imported (`:829`).

cc @n-gao

Issue #175, opened 2026-06-04
Lens[S, R] is declared covariant in R — unsound (R in set input position)


jonkhler:
`src/kups/core/lens.py:128-130`
```python
S = TypeVar("S", covariant=True)
R = TypeVar("R", covariant=True)
R2 = TypeVar("R2", covariant=True)
```
`R` appears in `set`'s input position, so `Lens` must be **invariant** in `R`. Covariance lets a `Lens[S, Sub]` stand in for `Lens[S, Super]`, after which `set` accepts a wrong `Super`.

## Repro (pyright: 0 errors)
```python
@dataclass
class Animal: legs: int
@dataclass
class Dog(Animal): ...
@dataclass
class Cat(Animal): ...
@dataclass
class Kennel: occupant: Dog

dog_lens = lens(lambda k: k.occupant, cls=Kennel)   # Lens[Kennel, Dog]
animal_lens: Lens[Kennel, Animal] = dog_lens         # accepted; should error
animal_lens.set(kennel, Cat(legs=4))                 # Cat written into a Dog slot, no diagnostic
```

## Why this is not a quick fix
Flipping `R`/`S` to invariant locally:
- the repro upcast errors as intended (`reportAssignmentType: R is invariant`);
- **49 errors** surface across `src/`, all subtype-focus lenses passed to general-focus params. Example, `src/kups/md/integrators.py:926`: `Lens[State, Table[ParticleId, _BarostatParticleData]]` into a param typed `Lens[State, Table[ParticleId, IsMomentumStepData]]`.

Each of the 49 is one of:
- **read-only** through the lens → sound fix is to type the param `View[S, T]` (covariant is correct for a getter);
- **read-write** → genuinely needs invariance, and covariance is currently masking an unsound write (latent bug).

Distinguishing requires per-site inspection. Naive removal of `covariant=True` also raises 3 `reportInvalidTypeVarUse` warnings (the protocol methods rebind `self: Lens[S2, R2]`, so the class vars read as covariant), so the real fix is PEP-695 inline variance + a `View` migration, not a one-liner.

The principled end state already exists in the module: `View` covariant (read-only), `Update` contravariant (write-only), `Lens` invariant.

## Not related
The `# type: ignore[reportIncompatibleMethodOverride]` on `MergedLens.set` (`lens.py:511`) is **not** a variance symptom; it persists under invariance (verified). It's a generic-method-override inference issue, as the inline comment already states. Don't conflate the two.

## Status
[...]

PR #221, opened 2026-06-26
refactor(lens): naming + docstring cleanups (issue #176)

sridhar-3009:
## Summary

Addresses all items from #176. Each change is independent and motivated by that issue.

### Changes

| Item | Change |
|------|--------|
| `apply` → `modify` | Rename on all four `Lens`/`BoundLens` protocol methods and ~20 call sites across `src/`. Matches python-lenses convention (`modify` / `over`); `apply` read ambiguously as "invoke the lens". |
| `LambdaLens` → `ExplicitLens` | The class wraps explicit named getter+setter functions — it is not specifically about lambdas. Rename is local to `lens.py` and `test/core/test_lens.py`. |
| `all_where_lens` docstring | `target_cls` was documented as `(currently unused)` and pyright flagged it. It is not dead: it binds the `Target` typevar for type inference (same pattern as `_cls` in `lens()` at :766). Fixed docstring to say "type-inference only, discarded at runtime" and added `_ = target_cls`. |
| `conditional` → `predicate` | Renamed in `all_where_lens` signature and body. The parameter is a boolean predicate; `where` already means "getter" elsewhere in the module. |
| `_NOT_SET_TYPE.OBJ` → `NOT_SET` | Member name `OBJ` was meaningless in context. |
| `bind()` no-`where` branch | Replaced `lambda x: x` with `identity`, which is already imported from `kups.core.utils.functools`. |

## Out of scope (per #176)

- `SimpleLens` rename (15+ call sites in `src/`)
- Lens variance fix (separate issue #175)

## Test plan

- [x] `python3.12 -c "import ast; ast.parse(...)"` — syntax valid
- [x] All 20 `apply` call sites renamed; `grep` confirms zero residual `.apply(` in `src/` and `test/`

### DIFF
--- a/src/kups/application/mcmc/data.py
+++ b/src/kups/application/mcmc/data.py
@@ -382,7 +382,7 @@ def mcmc_state_from_config(
     n_host = len(hp.keys)
     if host.cell_replication is not None:
         pos_lens = lens(lambda x: x.positions, cls=Particles)
-        p = pos_lens.apply(p, cell.wrap)
+        p = pos_lens.modify(p, cell.wrap)
         cell, p = make_supercell(cell, host.cell_replication, p, pos_lens)
         n_host = len(p.positions)
     n_ads = len(adsorbates)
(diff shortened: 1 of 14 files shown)


n-gao:
Hi @sridhar-3009, thanks a lot for your contribution! Would you mind resolving the merge conflicts?

Thanks a lot!

PR #225, opened 2026-06-29
feat(core): domain-decomposed potentials under shard_map

jonkhler:
Run classical potentials under `shard_map` domain decomposition, replicate-all: every device holds all N atoms, `origin: Index[OriginDeviceId]` marks ownership, and each device builds only the edges incident on its **owned** atoms by passing them as the neighbor list's `queried_keys` — owned atoms drive the cell-list stencil, so the costly candidate/distance work is O(N/D) while the key side is cheap-hashed. No ghost table: the owned-incident edges (global ids + image shifts) *are* the device's graph.

Correctness lives in the reduction, not the edge set. The stock energy functions reduce owned-only through the `Decomposition` seam (one PR below): `Sharded` masks non-owned rows and per-device energies come out as per-system **partials**; `ShardedPotential` psums ONLY the energy — gradients of the replicated positions/cell are already mesh-summed by `shard_map`'s transpose, and psumming them again would multiply forces by the device count (pinned by a unit test). The Ewald reciprocal term combines its owned partial structure factor across the mesh *inside* the differentiated energy (two PRs below).

Contents:
- `core/domain.py`: `Sharded`, `partition_equal_counts` (equal-count contiguous chunks in table order — ownership never affects correctness, only query locality; a space-filling-curve partitioner is the documented upgrade path if profiling shows locality matters), `owned_subset`, `sharded_local_edges`, `ShardedRadiusGraphConstructor` (no incremental path — a patch raises; static guard that origin labels match the mesh axis size).
- `core/sharding.py`: the mesh axis-name convention + replicated `device_put`. (jax's sentinel-unflatten round-trips need no per-site guards here: the placeholder detection of #290, at the bottom of the stack, covers them.)
- `owned_subset` records its capacity requirement as the max owned count over **all** devices (device-invariant, computed from the replicated origin labels), so the assertion leaves a `shard_map` through the default replicated assertion context and overflow **raises** instead of silently dropping atoms (pinned by test).
- DD gates on the 4-device simulated CPU mesh (`XLA_FLAGS` env set by #291): LJ energy+forces and the Ewald reciprocal term (with **net charge**) match single-device to machine precision. The segment-primitive vma rules the transpose relies on live in #291.

## Speedup (2× L4, strong scaling)

Measured with the stack-top drivers (#260) at 13 824 atoms, cutoff-swept coordination, marginal per-step cost (compile and warmup excluded):

| neighbors/atom | MD ms/step (D=1 → D=2) | MD speedup | relax ms/step (D=1 → D=2) | relax speedup |
|---|---|---|---|---|
| ~7  | 2.91 → 1.88 | **1.55×** | 6.6 → 6.0 | 1.10× |
| ~16 | 4.67 → 2.67 | **1.75×** | 8.3 → 6.7 | 1.23× |
| ~33 | 8.40 → 3.99 | **2.10×** | 9.9 → 8.0 | 1.24× |
| ~77 (realistic dense matter) | 28.2 → 13.7 | **2.05×** | 29.0 → 16.6 | 1.75× |

MD steps run in fused blocks of 10; relaxation pays a per-step host sync for the convergence check, which caps its scaling at low coordination. Expensive potentials (MLIAP) would dominate the step further, pushing end-to-end scaling toward the force-kernel limit at lower coordination. Measured at stack-top commit `25c43c5`.

Trajectories are bit-identical to single-device runs (gated in CI on the simulated mesh, and by the D=1 vs D=2 app-level trajectory tests).

**Scope:** `EwaldLongRangeComposer` / `evaluate_ewald_potential` still build `Replicated` point clouds and MLIAP/torch potentials are not decomposition-aware — DD covers the classical potentials via the graph constructor and direct energy calls. The runnable MD/relaxation drivers sit one PR above (#260).

Stack: #290 (pytree placeholder detection) → #291 (vma shim) → #258 (reduction seam) → #259 (per-atom Ewald) → **this PR** → #260 (drivers). (#257, the queried-keys self-image fix, is merged.)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,6 +124,7 @@ conflicts = [[{ extra = "uma" }, { extra = "mace" }]]
 
 [tool.pytest_env]
 JAX_ENABLE_X64 = "True"
+XLA_FLAGS = { value = "--xla_force_host_platform_device_count=4", "skip_if_set" = true }
 JAX_PLATFORMS = { value = "cpu", "skip_if_set" = true }
 JAX_DISABLE_MOST_OPTIMIZATIONS = { value = "True", "skip_if_set" = true }
 XLA_FLAGS = { value = "--xla_force_host_platform_device_count=4", "skip_if_set" = true }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -124,7 +124,6 @@ conflicts = [[{ extra = "uma" }, { extra = "mace" }]]
 
 [tool.pytest_env]
 JAX_ENABLE_X64 = "True"
-XLA_FLAGS = { value = "--xla_force_host_platform_device_count=4", "skip_if_set" = true }
 JAX_PLATFORMS = { value = "cpu", "skip_if_set" = true }
 JAX_DISABLE_MOST_OPTIMIZATIONS = { value = "True", "skip_if_set" = true }
 XLA_FLAGS = { value = "--xla_force_host_platform_device_count=4", "skip_if_set" = true }
--- a/src/kups/core/domain.py
+++ b/src/kups/core/domain.py
@@ -23,6 +23,7 @@ from typing import Any, Literal, Protocol, override, runtime_checkable
 
 import jax
 import jax.numpy as jnp
+import numpy as np
 from jax import Array
 
 from kups.core.capacity import Capacity
@@ -36,14 +37,13 @@ from kups.core.sharding import shard_axis
 from kups.core.typing import (
     HasCell,
     HasOrigin,
-    HasPositionsAndSystemIndex,
     IsState,
     OriginDeviceId,
     ParticleId,
     SystemId,
 )
 from kups.core.utils.jax import dataclass, field
-from kups.core.utils.math import triangular_3x3_matmul
+from kups.core.utils.ops import where_broadcast_last
 from kups.potential.common.graph import Decomposition, HyperGraph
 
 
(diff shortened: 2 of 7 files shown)


jonkhler:
## Design reasoning

**Replicate-all + AD.** Every device holds all atoms; `origin` is a reduction *label*, not a spatial region. Each device builds owned-incident edges (owned `rh` vs all `lh`), reduces energy owned-only, and psums it. Forces/cell-grad are cotangents of the replicated input, so `shard_map`'s transpose mesh-sums them for free — **psum the energy, never the forces**.

**Migration-free by construction.** Because data is replicated, an owned atom's full neighborhood is always local no matter how it moves or who owns the neighbors; ownership is a fixed label and energy is partition-independent. So no ghost exchange and no atom migration are needed for correctness (unlike LAMMPS, where each rank stores only its subdomain). Pinned by `test_domain_dynamics`: multi-step MD/relax trajectory == single-device as atoms move.

**Layer-0 is near-optimal in the no-comm regime.** The cell list already restricts the expensive distance/force work to O(N/D); only the cell-hash is O(N), and the bench shows it isn't the bottleneck (~1.9× strong scaling on 2 GPUs at realistic coordination ~50–70 nbrs/atom). Tightening the halo buys nothing without sharding — so this PR deliberately ships the simplest correct decomposition.

## What it would take to go further (region-DD / memory scaling)

Tracked in `plans/region_dd.md`. Not in this PR; needed only when N must exceed one device's memory.

- **Region ownership:** `origin = Partitioner(positions)` recomputed per replan; `Repartitioner` becomes the replan operator (present here, kept as the seam, then load-bearing + tested).
- **Halo as index set:** `lh = owned ∪ halo` (cutoff+skin shell from the cell list) + an lh→global remap; `cap_owned`/`cap_halo`.
- **Amortize on #194** (Verlet-skin rebuild trigger): replan fires on reneighbor.
- **The actual win needs one of:** region-restricted hashing (avoid the O(N) hash — SFC-key shortcut is unsound, needs care) **or** sharded data + gather-by-owner comm (the real O(N/D)-memory prize). Replicated + no-comm region-DD is interface plumbing only.

The `Partitioner -> Index[OriginDeviceId]` interface already admits arbitrary schemes (RCB/tiled, Hilbert, weighted); coherence affects halo *size*, not correctness — and there's no LAMMPS-style `brick`/`tiled` comm-style split because there's no comm.

jonkhler:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/225?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#260** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/260?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#225** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/225?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/225?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#259** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#258** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#291** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/291?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#290** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/290?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

PR #242, opened 2026-07-21
perf: allow skipping redundant query-cell dedup

jonkhler:
## Summary

- add an opt-in `promise_unique_cells=True` path to `CellListSelector` (named as a caller promise about the data, following the pattern of JAX's `promise_in_bounds`)
- replace the multi-key `jnp.unique` sort with the single-key hash argsort in `subselect` when callers prove stencil cells are distinct
- retain deduplication by default for backward compatibility (`promise_unique_cells=False`)

## Measurement

On one NVIDIA L4 for 8 independent 4,096-point periodic queries (884,736 expanded query-cell records), candidate generation falls from 6.751 ms to 5.415 ms. The 442,327-edge count, row/column checksums, cost sum, and capacity status are identical.

The intended caller guard requires at least three bins along each periodic axis; the option remains explicit because bin counts are runtime data.

## Validation

- focused cell-list and periodicity tests: passed
- Ruff format/check: passed
- Pyright: passed
- rebased onto current `main` to pick up the JAX 0.11 `scan` `num_consts` compatibility fix (#249) that was breaking the `highest` CI jobs

### DIFF
--- a/src/kups/core/neighborlist/cell_list.py
+++ b/src/kups/core/neighborlist/cell_list.py
@@ -40,7 +40,7 @@ from kups.core.neighborlist.types import (
     PipelineContext,
 )
 from kups.core.typing import ParticleId, SystemId
-from kups.core.utils.jax import dataclass, jit
+from kups.core.utils.jax import dataclass, field, jit
 
 
 class IsCellListParams(Protocol):
@@ -81,6 +81,7 @@ def _cell_list_subselect(
     cutoffs: Array,
     max_num_cells: Capacity[int],
     max_num_candidates: Capacity[int],
+    deduplicate_query_cells: bool = True,
 ) -> Candidates:
     cell = systems.data.cell
     key_positions, _ = cell.fold(keys.data.positions)
@@ -138,7 +139,7 @@ def _cell_list_subselect(
     # all-True, making the where a no-op.
     query_neighborhood_hashes = jnp.where(in_cell, hashes, cell_oob)
 
-    if not is_single_cell:
+    if not is_single_cell and deduplicate_query_cells:
         unique_queries = jnp.unique(
             jnp.stack([query_neighborhood_hashes, query_original.indices], axis=-1),
             axis=0,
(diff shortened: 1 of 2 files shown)


src/kups/core/neighborlist/cell_list.py
```diff
@@ -170,13 +171,17 @@ class CellListSelector:
...
+    be disabled when the caller guarantees at least three cells along every
+    periodic axis; this replaces the multi-key unique sort with a single-key
+    hash sort.
     """
 
     cutoffs: Table[SystemId, Array]
     max_cells: Capacity[int]
     max_candidates: Capacity[int]
     max_image_candidates: Capacity[int]
+    deduplicate_query_cells: bool = field(default=True, static=True)
```
n-gao: Should we name it like `promise_unique_cells`, like `promise_in_bounds` in https://docs.jax.dev/en/latest/_autosummary/jax.Array.at.html ? That way the user makes a statement about the data and the optimization can be done in the background.

jonkhler: yes sounds good - will change accordingly and pin gyou when updated

jonkhler: Done in 890906d — renamed to `promise_unique_cells` (default `False`), matching the `promise_in_bounds` naming pattern. Also rebased onto `main` so the `highest` CI jobs pick up the `num_consts` fix from #249. @n-gao ready for another look.

src/kups/core/neighborlist/cell_list.py
```diff
@@ -138,7 +139,7 @@ def _cell_list_subselect(
     # all-True, making the where a no-op.
     query_neighborhood_hashes = jnp.where(in_cell, hashes, cell_oob)
 
-    if not is_single_cell:
+    if not is_single_cell and not promise_unique_cells:
```
n-gao, line 154: do we want to add a runtime assertion? Not strictly necessary since we "promise" but probably doesn't really affect the runtime and would be safer?

n-gao (approved):
We can merge it as is or add the assertion as in the comment. I'll leave it up to you.

src/kups/core/neighborlist/cell_list.py
jonkhler: Good call — added in 398c0c7. The fast path now emits a `runtime_assert` that every periodic axis has at least three cells (non-periodic axes can't wrap, so they're exempt), with a test that a broken promise raises via `raise_assertion()` and a satisfied one doesn't. @n-gao could you take one more look / re-approve?

PR #250, opened 2026-08-03
fix(Table): Enforce sorted primary keys in `Table` and add `Table.new`

alexggmatthews:
`Table` documented its `keys` as "unique, sorted" but `__post_init__` only checked uniqueness. An unsorted table therefore constructed happily and then failed the moment it met `Index`, which does enforce sortedness -- `t.index` and `t[t.index]` both raised `ValueError: Keys must be unique and sorted.` from deep inside `Index.__post_init__`.

Enforce the invariant where it is documented:

- `_check_keys` walks the key tuple once and raises `ValueError` with separate messages for unsorted and duplicate keys, pointing the caller at `Table.new`. This replaces the previous `np.unique` uniqueness assert, so it is a `ValueError` rather than a stripped-under-`-O` assert, drops the `numpy` import, and is 4-9x faster than the check it replaces despite testing more (8.2 -> 0.9us at n=8; 5.15 -> 0.95ms at n=10,000). This matters because `__post_init__` re-runs on every pytree unflatten. It also fixes a latent bug in the old check: `np.asarray(keys, dtype=object)` builds a 2-D array for tuple-valued keys, which `np.unique` then miscounts.

- `Table.new(keys, data)` sorts keys and permutes `data` to match, mirroring `Index.new`. Rows are permuted via `bind(rows).at((order,)).get()` -- the same lens path `__getitem__` uses -- so nested `Index` foreign-key columns are permuted with the rows instead of being skipped as opaque leaves.

- `_validate_and_broadcast` is extracted as `_broadcast_rows` and shared, so `new` broadcasts size-1 leaves before permuting rather than gathering them out of bounds.

Instrumenting `__post_init__` across the suite found 7,443 multi-key `Table` constructions at 302 sites, of which exactly one was unsorted: a test that deliberately checks `Table.join` reordering. Nothing in `src/` constructs unsorted keys, so enforcement costs one existing test assertion, which now builds its input with `Table.new` and asserts the same result.

### DIFF
--- a/src/kups/core/data/table.py
+++ b/src/kups/core/data/table.py
@@ -19,7 +19,6 @@ from typing import (
 
 import jax
 import jax.numpy as jnp
-import numpy as np
 from jax import Array
 
 from kups.core.data.batched import Batched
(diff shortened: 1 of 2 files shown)


n-gao (approved):
LGTM

PR #17, opened 2026-04-21
feat: flat-histogram TMMC for adsorption (Witman 2018)

jonkhler:
## Summary

Top of the flat-histogram TMMC stack: the `kups_mcmc_nvtw` simulation entry point.

Fans out $N_\mathrm{max} + 1$ parallel NVT systems (one per macrostate $N$) over a shared host framework. Each cycle runs a displacement loop at fixed $N$, then ghost insertion/deletion probes accumulating the TMMC C-matrix and per-macrostate energy moments (Witman, Mahynski & Smit 2018). Cumulative accumulator snapshots are written to HDF5 per cycle; `summarize` bundles the final state into a `TMMCSummary` for isotherms / isosteric heats. Ewald and blocking spheres are auto-included from state, matching `kups_mcmc_widom`.

Also: TMMC HDF5 writer layout in `logging.py`, `docs/simulations.md` section, example config, and structural + end-to-end tests (macrostate populations, trial-count bookkeeping, $N=0$ deletion masking, finite $\ln Q_c$ reconstruction, isotherm bounds).

## Stack

Split bottom-up out of the original monolithic version of this PR (~2,500 lines); each layer is independently green:

1. #252 — fix: raise on unrecognised adsorbate symbols instead of deriving mass=0
2. #253 — feat: autodiff-safe cubic_roots and Peng-Robinson fugacity
3. #254 — feat: TMMC C-matrix and energy-moment accumulators
4. #255 — feat: flat-histogram (TMMC) post-processing
5. #256 — refactor: split mcmc_state_from_config into prepare_host + place_adsorbates
6. **#17 — this PR (top)**

## Notes vs. the original branch

- The plain-Widom portion already landed via #126 (with the review-flagged $q_\mathrm{st}$ / cell-volume bugs fixed there); the stack builds on it.
- `EwaldConfig.enabled` dropped: main auto-skips Ewald for neutral systems via `state.is_charged`.
- `md_nvtw` (hybrid MD + TMMC) is deferred to a follow-up — its N=5 cross-validation was outstanding in the original PR and the MD stack has since been reworked; the TMMC machinery here is entry-point-agnostic so it can slot in later.

## Test plan

- [x] Full suite on the combined tree: 1928 passed, 7 skipped (identical tree previously ran green through CI).
- [x] `pyrefly check src/`: 0 errors; pyright clean on changed files; ruff clean.
- [ ] GCMC cross-validation at production cycle counts before a release tag.

[...]

PR #254, opened 2026-08-05
feat: TMMC C-matrix and energy-moment accumulators (Witman 2018)

jonkhler:
Adds the two flat-histogram accumulators to `kups.mcmc.widom`, alongside the existing `WidomStatistics`:

- `TransitionStatistics`: collection-matrix accumulator for N → N±1 ghost moves (Witman 2018 eq 5–7). Deletion trials increment the trial count even at N=0 (zero accepted fraction) so P(0 → 1) is not inflated.
- `EnergyMoments` / `EnergyCumulants`: Pébay/Welford one-pass central moments 1–4 of the per-system potential energy, finalized into the β-derivative cumulants used by the ln Q_c Taylor expansion.

Tests: clamping/masking semantics, accumulation over repeated updates, Welford vs two-pass NumPy reference, per-system independence.

Part of the TMMC stack split out of #17.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #253, opened 2026-08-05
feat: autodiff-safe cubic_roots and Peng-Robinson fugacity

jonkhler:
- `custom_jvp` on `kups.core.utils.math.cubic_roots` via implicit differentiation (`dz/dp = -(∂F/∂p)/(∂F/∂z)`). Forward numerics unchanged (companion-matrix eigvals); only the pathological `eigvals` gradient is bypassed.
- Double-where NaN sanitisation in the Peng-Robinson fugacity root selection: unphysical roots get a safe placeholder Z that is never selected but keeps unused gradient branches finite.
- Regression tests: `jax.grad` of log-fugacity w.r.t. P and T is finite and matches central differences across gas / supercritical states.

Enables isosteric heats as a `jax.grad` through the loading function (no finite-difference hyperparameter) — consumed by the flat-histogram post-processing higher in the stack.

Part of the TMMC stack split out of #17.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #252, opened 2026-08-05
fix: raise on unrecognised adsorbate symbols instead of deriving mass=0

jonkhler:
United-atom pseudo-symbols (`CH4`, `CH3`, ...) and virtual sites (`Mw`, `M_n2`) are unknown to ASE, so the fallback silently produced `mass=0`. MC never notices, but anything dividing by mass (MD integrators, velocity initialisation) NaN-cascades.

Masses now derive from `atomic_numbers` (settable explicitly, with 0 marking a massless virtual site) rather than being hand-written; symbols ASE cannot resolve raise a pydantic `ValidationError` unless `atomic_numbers` or `masses` are given. Example configs set `atomic_numbers`; the united-atom examples keep explicit masses since a pseudo-atom mass (e.g. CH4 = C + 4H = 16.043) is not derivable from any Z. An after-validator additionally rejects per-atom fields whose length doesn't match `symbols`.

Bottom of the TMMC stack split out of #17.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #255, opened 2026-08-05
feat: flat-histogram (TMMC) post-processing

jonkhler:
`kups.mcmc.flat_histogram`: pure-function post-processing for TMMC adsorption data —

- Reconstructs `ln Q_c(N, V, β_sim)` from the C-matrix as a single `cumsum` (Witman 2018 eq 7–8).
- Taylor-extrapolates in β using the collected energy cumulants (eq 9–10).
- Derives adsorption isotherms, isosteric heats (Clausius–Clapeyron via `jax.grad` through the Peng-Robinson EOS — no `delta_T` hyperparameter; relies on the autodiff-safe `cubic_roots` lower in the stack), and working capacities.
- `TMMCSummary` bundles the reconstruction + EOS parameters for one-call post-hoc use.

Tests: transition-probability normalisation, Langmuir analytical round-trip to machine precision, extrapolation order behaviour, isosteric heat vs central differences, isotherm monotonicity. Everything is vectorised over pressure grids; no entry-point coupling.

Part of the TMMC stack split out of #17.

### DIFF
--- a/src/kups/mcmc/flat_histogram.py
+++ b/src/kups/mcmc/flat_histogram.py
@@ -1,0 +1,558 @@
+# Copyright 2024-2026 Cusp AI
+# SPDX-License-Identifier: Apache-2.0
+
+r"""Post-processing for flat-histogram (TMMC) adsorption simulations.
+
+Implements the temperature-extrapolation scheme of Witman, Mahynski & Smit
+(2018) [^witman2018] for computing adsorption isotherms, isosteric heats and
+working capacities from an NVT+W simulation performed at a single temperature.
+
+Pipeline:
+
+1. C-matrix $\to$ transition probabilities $P(N \to N \pm 1)$ (eq 7).
+2. Transition probabilities $\to$ $\ln Q_c(N, V, \beta_\mathrm{sim})$ via
+   a prefix sum of the detailed-balance ratio (eq 8). This replaces a
+   Python loop with a single [`jnp.cumsum`][jax.numpy.cumsum].
+3. Taylor-expand $\ln Q_c(N, V, \beta)$ in $\beta$ using the per-macrostate
+   energy cumulants collected during simulation (eq 9--10).
+4. Reweight into the grand canonical ensemble (eq 3) using
+   [peng_robinson_log_fugacity][kups.mcmc.fugacity.peng_robinson_log_fugacity]
+   for the reservoir fugacity.
+
+The isosteric heat of adsorption (eq 19) is computed with [`jax.grad`][jax.grad]
+through the full $\langle N \rangle(T, \ln P)$ pipeline, including the
+Peng--Robinson equation of state. Autodiff is enabled by the implicit-function
+JVP attached to [`cubic_roots`][kups.core.utils.math.cubic_roots] and the
+double-where sanitisation in
+[`peng_robinson_log_fugacity`][kups.mcmc.fugacity.peng_robinson_log_fugacity].
+No ``delta_T`` hyperparameter and no finite-difference truncation error.
+
+All public functions and the [`TMMCSummary`][kups.mcmc.flat_histogram.TMMCSummary]
+dataclass operate on plain [`jax.Array`][jax.Array] inputs and are fully
+vectorised across pressure grids.
+
+[^witman2018]:
+    Witman, M., Mahynski, N. A. & Smit, B. (2018). *J. Chem. Theory Comput.*
+    **14**, 6149--6158. DOI: 10.1021/acs.jctc.8b00534
+"""
+
+from __future__ import annotations
+
+import jax
+import jax.numpy as jnp
+from jax import Array
+
+from kups.core.constants import BOLTZMANN_CONSTANT, KELVIN, PASCAL
+from kups.core.utils.jax import dataclass, field
+from kups.mcmc.fugacity import peng_robinson_log_fugacity
+from kups.mcmc.widom import EnergyCumulants
+
+type LogPartitionFunction = Array
+r"""Natural logarithm of $Q_c(N, V, \beta)$, shape ``(n_max+1,)``."""
+
+type MacrostateDistribution = Array
+r"""Grand-canonical distribution $\Pi(N; \mu, V, \beta)$, shape ``(n_max+1,)``."""
+
+type Loading = Array
+r"""Expected loading $\langle N \rangle$, scalar or shape ``(n_pressures,)``."""
+
+type IsostericHeat = Array
+r"""Isosteric heat $q_\mathrm{st}$ [energy], scalar or shape ``(n_pressures,)``."""
+
+type InsertionProbability = Array
+r"""TMMC insertion probability $P(N \to N+1)$, shape ``(n_max+1,)``."""
+
+type DeletionProbability = Array
+r"""TMMC deletion probability $P(N \to N-1)$, shape ``(n_max+1,)``."""
+
+
+def transition_probabilities(
+    acceptance_insertion: Array,
+    acceptance_deletion: Array,
+    n_trials_insertion: Array,
+    n_trials_deletion: Array,
+) -> tuple[InsertionProbability, DeletionProbability]:
+    r"""Normalise accumulated C-matrix counts into per-macrostate transition probabilities.
+
+    Applies Witman 2018 eq 7:
+
+    $$P(N \to N + \Delta) = \frac{C(N, N + \Delta)}{\sum_\Delta C(N, N + \Delta)}.$$
+
+    Args:
+        acceptance_insertion: Accumulated $\min(1, \mathrm{acc})$ for
+            $N \to N+1$, shape ``(n_max+1,)``.
+        acceptance_deletion: Accumulated $\min(1, \mathrm{acc})$ for
+            $N \to N-1$, shape ``(n_max+1,)``.
+        n_trials_insertion: Ghost-insertion trial count per macrostate.
+        n_trials_deletion: Ghost-deletion trial count per macrostate
+            (includes trials at $N = 0$, which contribute zero acceptance).
+
+    Returns:
+        A pair ``(P(N → N+1), P(N → N-1))``, each shape ``(n_max+1,)``.
+    """
+    total = n_trials_insertion + n_trials_deletion
+    safe_total = jnp.where(total > 0, total, 1)
+    return (
+        acceptance_insertion / safe_total,
+        acceptance_deletion / safe_total,
+    )
+
+
+def reconstruct_log_partition_fn(
+    insertion_probability: InsertionProbability,
+    deletion_probability: DeletionProbability,
+    beta: Array,
+    log_fugacity: Array,
+) -> LogPartitionFunction:
+    r"""Reconstruct $\ln Q_c(N, V, \beta_\mathrm{sim})$ from TMMC transition probabilities.
+
+    Applies Witman 2018 eq 8 as a prefix sum over macrostates:
+
+    $$\ln Q_c(N+1) - \ln Q_c(N)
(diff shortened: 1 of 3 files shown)


jonkhler:
**TMMC stack** (merge bottom-up; each layer independently green):
1. #252 — fix: adsorbate mass raise
2. #253 — feat: Peng-Robinson autodiff
3. #254 — feat: TMMC accumulators
4. #255 — feat: flat-histogram post-processing
5. #256 — refactor: prepare_host / place_adsorbates split
6. #17 — feat: kups_mcmc_nvtw entry point (top)

jonkhler:
* **#17** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/17?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#256** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/256?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#255** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/255?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/255?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#254** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/254?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#253** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/253?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#252** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/252?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+r"""Isosteric heat $q_\mathrm{st}$ [energy], scalar or shape ``(n_pressures,)``."""
+
+type InsertionProbability = Array
+r"""TMMC insertion probability $P(N \to N+1)$, shape ``(n_max+1,)``."""
+
+type DeletionProbability = Array
+r"""TMMC deletion probability $P(N \to N-1)$, shape ``(n_max+1,)``."""
+
+
+def transition_probabilities(
```
n-gao, line 69: Could be a method of TransitionStatistics or accept TransitionStatistics? At least it has all the same fields.

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+    acceptance_insertion: Array,
+    acceptance_deletion: Array,
+    n_trials_insertion: Array,
+    n_trials_deletion: Array,
+) -> tuple[InsertionProbability, DeletionProbability]:
```
n-gao, line 74: A NamedTuple here?

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+        beta: Simulation inverse temperature $1/(k_B T_\mathrm{sim})$
+            [1/energy].
+        log_fugacity: Simulation log-fugacity $\ln f$ [dimensionless after
+            unit scaling].
+
+    Returns:
+        $\ln Q_c(N, V, \beta_\mathrm{sim})$ of shape ``(n_max+1,)``, anchored
+        to ``0`` at $N = 0$.
+    """
+    ratio = jnp.log(insertion_probability[:-1]) - jnp.log(deletion_probability[1:])
```
n-gao, line 131: log_ratio

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+
+    db = beta_target - beta_sim
+    result = log_partition_fn_sim - cumulants.mean * db
+    if order >= 2:
+        result = result + cumulants.variance * db**2 / 2.0
+    if order >= 3:
+        result = result + cumulants.third * db**3 / 6.0
+    if order >= 4:
+        result = result + cumulants.fourth * db**4 / 24.0
+    return result
```
n-gao, line 187: this could be written more generically for arbitrary degrees

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+
+def isotherm(
+    log_partition_fn: LogPartitionFunction,
+    beta: Array,
+    pressures: Array,
+    temperature: Array,
+    critical_pressure: float,
+    critical_temperature: float,
+    acentric_factor: float,
+) -> Loading:
```
n-gao, line 255: should we name these more specifically? I could get an isotherm also from gcmc simulations?

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+
+def isosteric_heat(
+    log_partition_fn_sim: LogPartitionFunction,
+    cumulants: EnergyCumulants,
+    beta_sim: Array,
+    pressures: Array,
+    temperature: Array,
+    critical_pressure: float,
+    critical_temperature: float,
+    acentric_factor: float,
```
n-gao, line 333: Any particular reason these must be floats?

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+    cumulants: EnergyCumulants,
+    beta_sim: Array,
+    temperature_ads: Array,
+    pressure_ads: Array,
+    temperature_des: Array,
+    pressure_des: Array,
+    critical_pressure: float,
+    critical_temperature: float,
+    acentric_factor: float,
+    order: int = 3,
```
n-gao, line 403: Should we group things in Tables? Also to indicate what is per system and what's per adsorbate?

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,558 @@
...
+def isosteric_heat(
+    log_partition_fn_sim: LogPartitionFunction,
+    cumulants: EnergyCumulants,
+    beta_sim: Array,
+    pressures: Array,
+    temperature: Array,
+    critical_pressure: float,
+    critical_temperature: float,
+    acentric_factor: float,
+    order: int = 3,
```
n-gao, line 334: This assumes we have a single adsorbate? Does it work for multiple systems?

n-gao (changes requested):
Not sure where to put this comment exactly, but it would help a lot if we use Tables to a) inform the cardinality of a quantity, i.e., per system, per adsorbate, etc. It also ensures that the shapes are correct.  
  
Second, this seems to assume that we have a single adsorbate which makes sense for an isotherm. Theoretically, I can do this in a matrix/tensor fashion with more kinds of adsorbates, right? Should we directly implement the more generic algorithm?

jonkhler:
I think both changes make sense! I will look into it!

src/kups/mcmc/flat_histogram.py
jonkhler: Done in 69cb9b1 — `transition_probabilities` now takes a `TransitionStatistics` directly (and `TMMCSummary.from_transition_statistics` likewise). Kept it a free function in this module rather than a method so widom.py stays accumulation-only and post-processing lives here.

src/kups/mcmc/flat_histogram.py
jonkhler: Done in 69cb9b1 — returns a `TransitionProbabilities` NamedTuple (`insertion`/`deletion`).

src/kups/mcmc/flat_histogram.py
jonkhler: Done in 69cb9b1.

src/kups/mcmc/flat_histogram.py
jonkhler: Done in 69cb9b1 — the expansion is now the generic sum lnQ_c(β') = lnQ_c(β) + Σ_{k=1}^{K} (−δβ)^k κ_k / k!, valid to whatever order the cumulants provide (`EnergyMoments` upstack now accumulates arbitrary-order moments). A quintic-lnQ_c exactness test at order 5 covers the generic path.

src/kups/mcmc/flat_histogram.py
jonkhler: Done in 69cb9b1 — the free functions are now `isotherm_from_log_partition_fn` / `isosteric_heat_from_log_partition_fn` / `working_capacity_from_log_partition_fn`, which says exactly what they consume (a GCMC-sampled isotherm never goes through lnQ_c). The `TMMCSummary` methods keep the short names since the class provides the context.

src/kups/mcmc/flat_histogram.py
jonkhler: No reason — loosened to `Array | float` in 69cb9b1, and `AdsorbateEOS` fields are now pytree leaves instead of static floats, so they can be traced (e.g. differentiating observables w.r.t. EOS parameters) or batched without recompilation.

src/kups/mcmc/flat_histogram.py
jonkhler: Partially — see the top-level reply on Tables. In 69cb9b1 the per-macrostate vs. per-system vs. per-species cardinality is now stated in every signature docstring and the module scope note; grouping into `Table`s here would need a new macrostate key type, which I'd rather introduce where the accumulators live in simulation state (#17) than in this pure post-processing module.

src/kups/mcmc/flat_histogram.py
jonkhler: Yes — single adsorbate species, and the arrays are one system's macrostate axis; batches of systems work by `vmap` over a leading axis. 69cb9b1 states this scope explicitly in the module docstring. True multi-component TMMC needs a multi-dimensional macrostate lattice (see top-level reply).

jonkhler:
Re the two design questions in the review:

**Tables for cardinality** — agreed on the goal. The catch in this module: every quantity here is indexed by the macrostate ($N$ = loading), which has no key type in `kups.core.data` yet — `SystemId`/`MotifId` tables only appear once the accumulators sit inside simulation state. What I did in 69cb9b1: every signature now documents its cardinality explicitly and the module docstring states the scope (per-macrostate arrays, one system, one species). I'd propose introducing a `MacrostateId`-keyed `Table` (and `Table[SystemId, TransitionStatistics]` state wiring) in #17 where the accumulators actually live in state — happy to do it there if you agree.

**Generic multi-adsorbate algorithm** — in principle yes: the macrostate becomes a lattice $(N_1, \ldots, N_s)$, the C-matrix gains an insertion/deletion row per species, and eq 8 turns into a lattice reconstruction where the prefix-sum along one axis is no longer well-defined under sampling noise (path-dependence; typically resolved by least-squares on the full transition graph rather than a cumsum). That's a substantially different algorithm than Witman 2018 (which is single-component), so I'd keep this PR 1-D and track the multi-component generalisation as a follow-up issue rather than block the stack on it. The single-species scope is now documented up front.

src/kups/mcmc/flat_histogram.py
```diff
@@ -0,0 +1,576 @@
+# Copyright 2024-2026 Cusp AI
+# SPDX-License-Identifier: Apache-2.0
+
+r"""Post-processing for flat-histogram (TMMC) adsorption simulations.
```
n-gao, line 4: these docstrings are just to long, one cannot read those.

n-gao (changes requested):
The docstrings are getting out of control ^^.   
  
I would still favor grouping in Tables if we need to refactor other parts to accommodate that then that's fine with me. We should be pedantic about how we store data.

PR #257, opened 2026-08-05
fix: count self-image pairs once in queried-keys neighbor lists

jonkhler:
A self-image candidate `(i, i, s)` is its own pair with reverse `(i, i, -s)`, and periodic image replication emits both. `QueriedKeysDedupMask`'s `key >= query` orientation rule kept **every** self-image copy, and `MirrorPairEdges` then emitted each one a second time: self-image interactions were double-counted whenever the cutoff exceeds a perpendicular cell length. Only queried-keys incremental updates were affected (full self-graphs never enter the dedup path); reproduced as a 1.3 % energy error on a 2-atom cell with cutoff > box.

The fix keeps exactly the lexicographically positive shift of a self-image row, so the mirror restores its negation once. The zero-shift self row is dropped (its mirror is itself; `ExclusionMask` drops it anyway as the pair's minimum image).

Regression: `queried_keys` over all ids must reproduce the full self-graph edge **multiset** exactly, pinned end-to-end for dense + cell-list selectors at deep replication (`ratio > 1`), plus a unit pin on the mask rule. Both fail on the previous code.

Found while stacking the domain-decomposition work (#225), which makes this path load-bearing for its primary energy.

Bottom of the domain-decomposition stack.

### DIFF
--- a/src/kups/core/neighborlist/masks.py
+++ b/src/kups/core/neighborlist/masks.py
@@ -73,9 +73,16 @@ class QueriedKeysDedupMask:
 
     Pair selectors emit candidates in ``keys`` space. When ``ctx.queried_keys``
     is set, the query side was restricted to those affected ``keys`` rows.
-    We keep edges whose key endpoint is unaffected, plus one orientation for
-    edges where both endpoints are affected. ``MirrorPairEdges`` restores the
-    reverse orientation after compaction.
+    We keep edges whose key endpoint is unaffected, plus ONE orientation of
+    every edge whose endpoints are both affected; ``MirrorPairEdges`` restores
+    the reverse orientation after compaction. For distinct endpoints the kept
+    orientation is ``key > query``. A self-image row ``(i, i, s)`` is its own
+    pair with reverse ``(i, i, -s)`` — both are emitted by image replication —
+    so the kept orientation is the lexicographically positive shift; keeping
+    both (as ``key >= query`` did) made the mirror emit each self-image twice
+    whenever the cutoff exceeds a perpendicular cell length. The zero-shift
+    self row is dropped here (its mirror is itself; ``ExclusionMask`` drops it
+    anyway as the pair's minimum image).
 
     Returns all-True for full self-graphs and bipartite queries.
     """
@@ -85,8 +92,16 @@ class QueriedKeysDedupMask:
     ) -> Array:
         if ctx.queried_keys is None:
             return jnp.ones((batch.key_idx.size,), dtype=bool)
-        return ~isin(batch.key_idx.indices, ctx.queried_keys, ctx.keys.size) | (
-            batch.key_idx.indices >= batch.query_idx.indices
+        key = batch.key_idx.indices
+        query = batch.query_idx.indices
+        shift = batch.edges.shifts[:, 0, :]
+        first_nonzero = jnp.take_along_axis(
+            shift, jnp.argmax(shift != 0, axis=-1)[:, None], axis=-1
+        )[:, 0]
+        return (
+            ~isin(key, ctx.queried_keys, ctx.keys.size)
+            | (key > query)
+            | ((key == query) & (first_nonzero > 0))
         )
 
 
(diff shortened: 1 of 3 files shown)


src/kups/core/neighborlist/masks.py
```diff
@@ -73,9 +73,16 @@ class QueriedKeysDedupMask:
...
+    We keep edges whose key endpoint is unaffected, plus ONE orientation of
+    every edge whose endpoints are both affected; ``MirrorPairEdges`` restores
+    the reverse orientation after compaction. For distinct endpoints the kept
+    orientation is ``key > query``. A self-image row ``(i, i, s)`` is its own
+    pair with reverse ``(i, i, -s)`` — both are emitted by image replication —
+    so the kept orientation is the lexicographically positive shift; keeping
+    both (as ``key >= query`` did) made the mirror emit each self-image twice
+    whenever the cutoff exceeds a perpendicular cell length. The zero-shift
+    self row is dropped here (its mirror is itself; ``ExclusionMask`` drops it
+    anyway as the pair's minimum image).
```
n-gao: to verbose

jonkhler:
* **#260** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/260?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#225** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/225?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#259** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#258** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/258?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#257** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/257?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/257?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

test/core/neighborlist/test_image_replication.py
```diff
@@ -883,6 +884,78 @@ def test_no_clearly_inside_neighbor_dropped_end_to_end(self, box, cutoff, xs):
...
+            f"self-graph: {dict((queried - full) + (full - queried))}"
+        )
+        # A self-image row (i, i, n) is in range iff |n @ A| < cutoff, so the
+        # regression is only exercised when the cutoff exceeds the shortest
+        # lattice vector (ortho_cubic @ 1.19 does; guard that it stays covered).
+        if cutoff > float(np.linalg.norm(np.asarray(matrix), axis=1).min()):
+            self_rows = [e for e in full if e[0] == e[1]]
+            assert self_rows, "case was expected to exercise self-image rows"
+
+
```
n-gao: can we make this simpler and with a single atom? It should also happen there?

n-gao: Can we also add a test that it works if the cutoff is > 2 * lattice length?

n-gao (approved):
Minor comments on style and test, otherwise LGTM.

src/kups/core/neighborlist/masks.py
jonkhler: Trimmed in 9782b4e to just the orientation rule (key > query for distinct endpoints, lexicographically positive shift for self-image rows); the key >= query regression history lives in the commit message and the mask unit test.

test/core/neighborlist/test_image_replication.py
jonkhler: Yes — it reproduces with a single atom. In 9782b4e the test is one atom in a 4 Å cubic box checked against the exact analytic self-image multiset (each shift with |n|·box < cutoff exactly once), for both full and queried-keys runs; ratio 2.3 also covers cutoffs above twice the lattice length. Verified all four parametrizations fail with the fix reverted.

PR #263, opened 2026-08-05
feat: add the deformation-aware Verlet-skin margin bound

jonkhler:
**Stack (bottom → top):** #261 → #262 → #263 → #264 → #194

Part of the **Verlet-skin stack** (assert-repair tests → neighborlist_factory → this PR → skin machinery → MD wiring).

Pure functions for a Verlet-skin neighbor list — no state, no steps, just the math:

- `effective_build_radii`: per-system build radius `cutoff + skin`, clamped to the single-image limit (half the smallest perpendicular length over periodic axes; no cap in vacuum). Refine-based reuse can only represent one periodic image per stored pair, so a compressing cell degrades to a thinner effective skin instead of a hard failure.
- `skin_margin`: the completeness margin of a list built at that radius,

  ```
  2·max|u|  +  r_build · max(0, 1 − σ_min(F))   ≤   r_build − cutoff
  ```

  with `F = h_ref⁻¹·h_now` the per-system cell deformation since the build (row convention: pair vectors map as `d ↦ d·F`, and so do the image lattice vectors) and `u` the minimum-image **non-affine** residual displacement `x − x_ref·F`. Any image of a non-listed pair now sits at distance ≥ `σ_min·r_build − 2·max|u|`, so the list is complete while the margin holds. Because `σ_min` sees the full deformation, **pure shear consumes margin like any other strain** (relevant for anisotropic NPT barostats) and expansion is correctly free. Accounting is fully per system: pairs never span systems, so one hot system neither charges nor triggers the others.

Tests: motion threshold at skin/2, compression fires / expansion doesn't, pure shear fires, triclinic boundary crossing measured exactly (a per-axis un-wrap would misread a 0.4 Å step as ~3.9 Å), non-periodic axes not un-wrapped, per-system accounting, and both clamp cases.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #262, opened 2026-08-05
feat: accept a neighborlist_factory in potential configs

jonkhler:
**Stack (bottom → top):** #261 → #262 → #263 → #264 → #194

Part of the **Verlet-skin stack** (assert-repair tests → this PR → margin bound → skin machinery → MD wiring).

All four `PotentialConfig.build` methods forward an optional `neighborlist_factory` (default `AdaptiveNeighborList.from_state`, behaviour unchanged) to their adapters, so a caller can inject an alternative neighbor-list construction — e.g. a stored, reused list — without touching the config schema. The LJ/torch/tojax adapters already accepted the parameter; this only threads it through.

Mechanical, default-preserving; no behavior change.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #261, opened 2026-08-05
test: pin assertion repair inside fused cond/loop propagators

jonkhler:
**Stack (bottom → top):** #261 → #262 → #263 → #264 → #194

Bottom of the **Verlet-skin stack** (this PR → neighborlist_factory → margin bound → skin machinery → MD wiring).

The runtime-assertion machinery composes `lax.cond`, `lax.while_loop` (`LoopPropagator`), `ResetOnErrorPropagator` and `propagate_and_fix`, but no test pinned the combination: an assert **with a fix** raised inside a cond branch inside a fused loop. The upcoming Verlet-skin neighbor list rests on exactly this composition, so pin the three behaviors first:

1. cond branches carrying assertions run on device inside the fused loop (`test_cond_rebuilds_inside_fused_loop`);
2. a mid-loop failure reverts the offending iteration and stalls the dispatch at the last valid state (the failed condition is state-deterministic, so later iterations re-fail and revert), and one fix + re-dispatch resumes from there — `max_tries=2` rules out implementations that discard the pre-failure work (`test_midblock_failure_reverts_fixes_and_resumes`);
3. an exception raised by a fix function propagates (`test_fix_exception_propagates`).

Pure test addition; no source changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #264, opened 2026-08-05
feat: add Verlet-skin build, reuse and rebuild machinery

jonkhler:
**Stack (bottom → top):** #261 → #262 → #263 → #264 → #194

Part of the **Verlet-skin stack** (assert-repair tests → neighborlist_factory → margin bound → this PR → MD wiring).

The stateful machinery on top of the margin bound:

- `build_skin_edges` — one conservative dense build at the clamped radius; `skin_neighborlist` — the cheap per-step reuse path (refine stored edges to the true cutoff, minimum-image shifts recomputed for the current cell).
- `RebuildSkinStep` — rebuilds when `state.should_rebuild` is set, via `lax.cond`, **on device**, so it fuses into blocked stepping. Only a cutoff that itself violates the single-image limit raises.
- `TriggerStep` — end of every step: sets the flag by extrapolating one step of per-system margin consumption, and hard-asserts no system's headroom went negative — the assertion is exactly the validity condition of that step's force eval. Runs inside the MD step's `ResetOnErrorPropagator` scope, so a step that outruns the margin is reverted and stepping stalls at the last valid state; the attached fix flips the flag and `propagate_and_fix` re-dispatches from there with a rebuild. The trigger formula is thereby a performance heuristic, not a correctness proof. A single step consuming more than a system's whole budget escalates with a configuration hint.
- **Capacity overflows self-repair including the stored edges**: the skin build's capacities sit behind a lens whose `set` also re-pads the stored edges to the grown build shape, so a retraced `lax.cond` keeps matching branch shapes.
- `seed_verlet_state` — the one sanctioned way to populate the Verlet fields. An eager seed build cannot surface capacity assertions (and silently outgrows undersized static params), so the seed refits the stored edges to the params-implied shape and schedules an assertion-covered rebuild for the first traced step.

Tests: refined stored list reproduces a fresh dense build exactly (cubic + triclinic), seeding contract (flag, copies, overgrown-seed refit), clamp vs fatal rebuild cases, and the backstop (fix flips flag; unabsorbable step escalates; healthy step extrapolates; two-system: only the exhausted system trips it, per-system headroom uncoupled).


jonkhler:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/264?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#194** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#264** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/264?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/264?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#263** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/263?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#262** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/262?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#261** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/261?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

src/kups/core/neighborlist/verlet.py
```diff
@@ -43,21 +43,94 @@
...
+assertion's fix — flip ``should_rebuild`` — and re-dispatches, so stepping
+resumes from the last valid state with a rebuild. (A fused iteration after the
+failure that satisfies its own margin — e.g. under fresh thermostat noise —
+commits soundly: the assertion *is* the validity condition for its step.) The
+trigger formula is thereby a performance heuristic, not a correctness proof:
+over-firing costs a rebuild, under-firing costs re-running a few steps. Only a
+single step that consumes more than a system's whole budget — unabsorbable by
+any rebuild schedule — escalates, with a configuration hint. The measurement
+point (end of step) trails the force evaluation by at most the intra-step
+position update, the standard LAMMPS-style slack the extrapolation absorbs.
```
n-gao: I dont think the method docstring should be in the module docstring

src/kups/core/neighborlist/verlet.py
```diff
@@ -43,21 +43,94 @@
...
+    @property
+    def stored_skin_edges(self) -> Edges[Literal[2]] | None: ...
+    @property
+    def reference_positions(self) -> Array | None: ...
+    @property
+    def reference_cell(self) -> Cell[AnyPeriodicity] | None: ...
+    @property
+    def should_rebuild(self) -> Array | None: ...
+    @property
+    def skin_headroom(self) -> Array | None: ...
```
n-gao: Can we group these like for the other neighborlists?

src/kups/core/neighborlist/verlet.py
```diff
@@ -82,6 +155,175 @@ def effective_build_radii(
...
+    one — mismatched candidate/image buffers trigger a pathological dense-build path).
+    Estimated at the unclamped radius, so a clamped build can only need less.
+    """
+    return UniversalNeighborlistParameters.estimate(
+        particles_per_system, systems, _skin_cutoffs(cutoffs, skin)
+    )
+
+
+def _fit_edges_like(
+    edges: Edges[Literal[2]], target: Any, invalid_index: int
```
n-gao: Avoid Any

n-gao: invalid_index = out_of_bounds_value ?

src/kups/core/neighborlist/verlet.py
```diff
@@ -43,21 +43,94 @@
...
+from kups.core.typing import IsState, SystemId
+from kups.core.utils.jax import dataclass, field
 
 # Above this ratio of build radius to a cell's perpendicular length, a build needs
 # more than one periodic image per pair and the refine-based reuse path is incomplete.
 _SINGLE_IMAGE_LIMIT = 0.5
 
 
+class IsVerletState(
+    IsState[Any, Any], IsNeighborListState[IsUniversalNeighborlistParams], Protocol
```
n-gao: The IsState is not necessary if it has IsNeighborListState.

src/kups/core/neighborlist/verlet.py
```diff
@@ -82,6 +155,175 @@ def effective_build_radii(
...
+            return leaf[:n]
+        fill = invalid_index if jnp.issubdtype(leaf.dtype, jnp.integer) else 0
+        pad = jnp.full((n - leaf.shape[0], *leaf.shape[1:]), fill, leaf.dtype)
+        return jnp.concatenate([leaf, pad], axis=0)
+
+    return jax.tree.map(fit, edges, target)
+
+
+@dataclass
+class _SkinParamsLens[State: IsVerletState](BaseLens[State, Any]):
```
n-gao: Avoid Any

src/kups/core/neighborlist/verlet.py
n-gao: We shouldn't have the arrays be optional None. It is better to have them Arrays and properly init them. If they switch from None -> Array, we have to do another compile pass (which can be lengthy).

src/kups/core/neighborlist/verlet.py
n-gao: If the fields cannot be None this special lense shouldnt be necessary.

src/kups/core/neighborlist/verlet.py
```diff
@@ -82,6 +155,175 @@ def effective_build_radii(
...
+            bind(state)
+            .focus(lambda s: s.stored_skin_edges)
+            .set(_fit_edges_like(state.stored_skin_edges, target, state.particles.size))
+        )
+
+
+def build_skin_edges[State: IsVerletState](
+    state: State, cutoffs: Table[SystemId, Array], skin: float
+) -> Edges[Literal[2]]:
+    """One conservative dense build at the clamped ``cutoff + skin`` radius.
```
n-gao: public methods should include Args and Return documentation.

src/kups/core/neighborlist/verlet.py
```diff
@@ -82,6 +155,175 @@ def effective_build_radii(
...
+    not surface — callers seeding a state eagerly must set
+    ``should_rebuild=True`` so the first traced step rebuilds under assertion
+    coverage. Used by
+    [`RebuildSkinStep`][kups.core.neighborlist.verlet.RebuildSkinStep].
+    """
+    cell = state.systems.data.cell
+    radii = effective_build_radii(
+        Table.broadcast_to(cutoffs, state.systems).data, skin, cell
+    )
+    skin_nl = DenseNearestNeighborList.new(
```
n-gao: this should be modular, could also be CellList

src/kups/core/neighborlist/verlet.py
n-gao: When grouped to a dataclass it could have a staticmethod for seeding them from the particles and system Tables before constructing the full state.

src/kups/core/neighborlist/verlet.py
```diff
@@ -82,6 +155,175 @@ def effective_build_radii(
...
+        cutoffs=cutoffs,
+    )
+
+
+def seed_verlet_state[State: IsVerletState](
+    state: State,
+    cutoffs: Table[SystemId, Array],
+    skin: float,
+    params: IsUniversalNeighborlistParams | None = None,
+) -> State:
```
n-gao: see comment above.

src/kups/core/neighborlist/verlet.py
```diff
@@ -125,3 +367,139 @@ def skin_margin(
     r_build = effective_build_radii(cutoffs, skin, cell_ref)
     consumed = 2.0 * u_max + r_build * jnp.maximum(0.0, 1.0 - sigma_min)
     return consumed, r_build - cutoffs
+
+
+def _flip_rebuild_flag[State: IsVerletState](
+    state: State, args: dict[str, Array]
+) -> State:
+    """Fix for an exhausted margin: request a rebuild and let the block replay.
```
n-gao: the args dict is very non kUPS.

src/kups/core/neighborlist/verlet.py
```diff
@@ -125,3 +367,139 @@ def skin_margin(
...
+            fmt_args={"headroom": headroom},
+            fix_fn=_flip_rebuild_flag,
+            fix_args={"deficit": single_step - budget},
+        )
+        flag = (headroom - single_step < 0.0).any()
+        return (
+            bind(state)
+            .focus(lambda s: (s.should_rebuild, s.skin_headroom))
+            .set((flag, headroom))
+        )
```
n-gao: Why It seems like extra steps to have two propagators where one flips a bit and marks the step as failed. The other then in the next iteration rebuilds the neighborlist. We should merge these into one which if the check fails (using jax.lax.cond) recomputes the full neighborlist and the fix function then just assigns the new edgelist. This also avoids having this bool in our State.

n-gao (changes requested):
There are some non-kUPSy parts and I think we can simplify the logic quite a bit.

src/kups/core/neighborlist/verlet.py
jonkhler: Moved to VerletSkinPropagator's class docstring; the module docstring is a short index now.

src/kups/core/neighborlist/verlet.py
jonkhler: Grouped into VerletSkinState with a seed(particles, systems, cutoffs, skin) classmethod. The arrays themselves are no longer Optional; only the single group field is, and it is set exactly once before the first trace, so the None→Array recompile can't occur mid-run. seed also passes particles.data.system.counts straight through — the previous rewrap would have silently relabelled the counts instead of letting Table.join catch a mismatch.

src/kups/core/neighborlist/verlet.py
jonkhler: All Anys removed — the lens is concretely typed against UniversalNeighborlistParameters and the shape target is the Edges pytree that eval_shape returns; also renamed invalid_index → out_of_bounds_value.

src/kups/core/neighborlist/verlet.py
jonkhler: Dropped. IsVerletState now extends only IsNeighborListState[IsUniversalNeighborlistParams], declaring the precisely-typed particles/systems tables the build actually reads.

src/kups/core/neighborlist/verlet.py
jonkhler: Gone with the same pass — no Any left in the module.

src/kups/core/neighborlist/verlet.py
jonkhler: Added Args/Returns to all public methods (estimate_skin_params, VerletSkinState.seed, build_skin_edges, skin_neighborlist, VerletSkinPropagator.__call__).

src/kups/core/neighborlist/verlet.py
jonkhler: Made injectable via a BackingNeighborListFactory protocol matching the concrete classes' .new classmethods — CellListNeighborList.new plugs straight in; the default is AdaptiveNeighborList.new, which already routes to the cell list for large systems.

src/kups/core/neighborlist/verlet.py
jonkhler: Addressed together with the grouped VerletSkinState above.

src/kups/core/neighborlist/verlet.py
jonkhler: Replaced with a plain deficit: Array fix argument.

src/kups/core/neighborlist/verlet.py
jonkhler: Merged into one VerletSkinPropagator wrapping the inner step, so there is a single propagator now. I kept the bit though: the fix runs eagerly, and an eager build outgrows the static params without recording the growth — refitting its output to the params-implied shape would silently drop edges (the case test_undersized_seed_capacities_are_repaired pins). The rebuild therefore runs in-trace under the lax.cond, where LensCapacity can grow the buffers, and the bit is the minimal channel from the eager fix to that traced rebuild. (Doing check+rebuild+re-step inside one cond would compile the force evaluation twice.) It's documented on the propagator, and the bit lives inside the grouped VerletSkinState rather than loose on the app state.

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+    """
+
+    neighborlist_params: UniversalNeighborlistParameters
+    edges: Edges[Literal[2]]
+    reference: SkinReference
+    should_rebuild: Array
+    headroom: Array
+
+    @classmethod
+    def seed(
```
n-gao, line 327: init

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+        if leaf.shape[0] >= n:
+            return leaf[:n]
+        fill = out_of_bounds_value if jnp.issubdtype(leaf.dtype, jnp.integer) else 0
+        pad = jnp.full((n - leaf.shape[0], *leaf.shape[1:]), fill, leaf.dtype)
+        return jnp.concatenate([leaf, pad], axis=0)
+
+    return jax.tree.map(fit, edges, target)
+
+
+def _build_edges[Holder](
```
n-gao, line 281: Holder -> State

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+        """
+        if params is None:
+            params = estimate_skin_params(
+                particles.data.system.counts, systems, cutoffs, skin
+            )
+        params_lens = identity_lens(UniversalNeighborlistParameters)
+
+        def build(
+            p: Table[ParticleId, NeighborListPoints],
+            s: Table[SystemId, NeighborListSystems],
```
n-gao, line 369: not sure why those need to be there, could be fully closed since it's only used to eval the shapes

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+    and the propagator in this module require it to be populated (see
+    [`VerletSkinState.seed`][kups.core.neighborlist.verlet.VerletSkinState.seed]).
+    """
+
+    @property
+    def particles(self) -> Table[ParticleId, NeighborListPoints]: ...
+    @property
+    def systems(self) -> Table[SystemId, NeighborListSystems]: ...
+    @property
+    def verlet_skin(self) -> VerletSkinState | None: ...
```
n-gao, line 402: Typically we don't allow for None

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+
+
+@dataclass
+class _SkinParamsLens[State: IsVerletState](
+    BaseLens[State, UniversalNeighborlistParameters]
+):
```
n-gao, line 408: this would be obsolete if verlet_skin would not be allowed to be None.

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+        target = jax.eval_shape(
+            lambda s: build_skin_edges(s, dummy_cutoffs, 0.0, self.neighborlist),
+            state,
+        )
+        edges = _fit_edges_like(skin_state.edges, target, state.particles.size)
+        return (
+            bind(state)
+            .focus(lambda s: s.verlet_skin)
+            .set(dataclasses.replace(skin_state, edges=edges))
+        )
```
n-gao, line 444: why with a dummy cutoff and not the real one?

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+
+    Args:
+        state: State carrying a seeded ``verlet_skin`` group.
+        cutoffs: True cutoffs (Å) per system.
+        skin: Requested skin width (Å).
+        neighborlist: Backing implementation performing the build.
+
+    Returns:
+        Conservative skin edges for the current positions and cells.
+    """
```
n-gao, line 472: redundant wrapper of `_build_edges`.

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+    (end of step) trails the force evaluation by at most the intra-step
+    position update, the standard LAMMPS-style slack the extrapolation
+    absorbs.
+
+    The rebuild request travels as the ``should_rebuild`` bit — rather than
+    the fix assigning a fresh edge list directly — because the build must run
+    *in-trace*: only there do its capacity assertions surface and auto-grow
+    the buffers, while an eager build inside the fix would outgrow the static
+    params without recording the growth, and refitting its output back to the
+    params-implied shape would silently drop edges.
```
n-gao, line 587: not the only one but the most outrageous one. Way to long docstring. One cannot read this.

src/kups/core/neighborlist/verlet.py
```diff
@@ -159,3 +218,475 @@ def skin_margin(
...
+            fix_args=single_step - margin.budget,
+        )
+        flag = (headroom - single_step < 0.0).any()
+        return (
+            bind(state)
+            .focus(lambda s: s.verlet_skin)
+            .set(
+                dataclasses.replace(skin_state, should_rebuild=flag, headroom=headroom)
+            )
+        )
```
n-gao, line 692: dataclasses.replace is not kUPSy.

n-gao (changes requested):
a lot of dataclasses.replace, use lenses instead. Also the docstrings are to long.

src/kups/core/neighborlist/verlet.py
```diff
@@ -53,6 +85,33 @@ def effective_build_radii(
...
+        systems: System table (cells) the builds will run in.
+        cutoffs: True cutoffs (Å) per system.
+        skin: Requested skin width (Å).
+
+    Returns:
+        Capacity hints for the skin build.
+    """
+    return UniversalNeighborlistParameters.estimate(
+        particles_per_system, systems, cutoffs.map_data(lambda c: c + skin)
+    )
```
n-gao, line 112: do we want this extra function that just adds something to the cutoff? What if we would have a different neighborlist that is not covered by the "Universal" part there. I would just call the original function with the enlarged cutoff

PR #265, opened 2026-08-05
feat: FFT-based smooth-PME reciprocal energy module

jonkhler:
**Stack (bottom → top):** #265 → #196

Bottom of the **PME stack** (this PR → PMESettings wiring).

## What

An **O(N log N) FFT-based smooth-PME reciprocal term** (Essmann et al. 1995) as a drop-in for `ewald_long_range_energy`, whose dense `(N, N_k, 2)` structure-factor tensor runs out of memory at large N. Charge spreading with cardinal B-splines, the Euler exponential-spline aliasing correction, and the same neutralizing-background term as direct Ewald so net-charged cells stay correct. Forces and the NPT stress come from autodiff, as for direct Ewald.

## Correctness fixes found while preparing this PR

Each verified test-first (fails against the previous code):

1. **Transposed fractional-coordinate / k-vector convention.** `frac = inv @ r` and `k = 2π·invᵀ·n` were both the transpose of this codebase's convention. The errors cancel in `k·r`, so PME matched Ewald on cubic cells — while computing the reciprocal energy of the *transposed lattice*: wrong energy for any triclinic cell, wrong off-diagonal `dE/dcell` for every cell (silently broken anisotropic NPT; `g[1,0] = -10.08` vs the correct `+0.42` on a cubic cell). Tests: `test_energy_matches_ewald_under_shear` (parametrized), `test_full_cell_gradient_matches_ewald` (full 3×3).
2. **Missing neutralizing-background term** for net-charged cells (all previous fixtures were neutral NaCl). Test: `test_net_charged_system_matches_ewald`.
3. **Per-system `alpha` via keyed lookup** (`alpha[pc.systems.index]`), so a batch whose parameter and system tables are keyed in different orders can't pick up the wrong splitting parameter.

## Removed relative to the original submission

The `fp32_reciprocal`/`pme_fp32` pipeline (zero callers/tests/benchmarks, threaded through 9 signatures) — re-add with a benchmark and a test pinning the accuracy claim if wanted.

## Validation

Tested against the direct-Ewald oracle for energy, forces, the full `dE/dcell` tensor, sheared cells, a net-charged cell, and a two-system batch. Tolerances `1e-5` (energy) / `atol=3e-4` (forces) — the old looser ones had 20–1500× headroom. Known open (documented in docstrings): `pme_mesh_for_cell` ignores `alpha`; deriving the mesh from `alpha` as standard PME codes do is future work.

### DIFF
--- a/src/kups/potential/classical/pme.py
+++ b/src/kups/potential/classical/pme.py
@@ -32,7 +32,7 @@ import numpy as np
 from jax import Array
 from jax.typing import DTypeLike
 
-from kups.core.cell import Periodic3D
+from kups.core.cell import Cell, Periodic3D
 from kups.core.data import Table, WithIndices
 from kups.core.lens import Lens, View, lens
 from kups.core.patch import IdPatch, Patch, Probe, WithPatch
@@ -62,8 +62,12 @@ class PMESettings:
     those ``None`` keeps the direct-Ewald reciprocal sum.
 
     Attributes:
-        mesh: Static FFT grid dimensions, held fixed across a trajectory (see
-            `pme_mesh_for_cell`).
+        mesh: Static FFT grid dimensions. Settings rather than state because FFT
+            shapes must be known at trace time, and shared by all systems in a
+            batch — one FFT over a ``(n_sys, *mesh)`` charge grid
+            (`pme_mesh_for_cell` sizes the mesh for the largest cell per axis).
+            Held fixed across a trajectory even as an NPT box fluctuates, as in
+            standard PME codes.
         order: Cardinal B-spline order.
     """
 
@@ -72,25 +76,23 @@ class PMESettings:
 
 
 def pme_mesh_for_cell(
-    cell_vectors: Array, spacing: float = 1.0, multiple_of: int = 2
+    cell: Cell[Periodic3D], spacing: float = 1.0, multiple_of: int = 2
 ) -> tuple[int, int, int]:
     """Pick static PME mesh dimensions for a reference cell (~``spacing`` Angstrom).
 
-    Resolution is measured across opposing faces, not along the lattice vectors, so
-    the spacing holds for sheared cells too. Dims are rounded up to a multiple of
-    ``multiple_of`` (cheaper FFTs). Use the *initial* cell; the mesh is held fixed
-    across an NPT trajectory.
+    Resolution is measured across opposing faces (``cell.perpendicular_lengths``),
+    not along the lattice vectors, so the spacing holds for sheared cells too. A
+    batched cell is reduced with a per-axis max, sizing the shared mesh for the
+    largest system. Dims are rounded up to a multiple of ``multiple_of`` (cheaper
+    FFTs). Runs on the host and returns Python ints because FFT shapes must be
+    static: call it at setup time with the *initial* cell; the mesh is then held
+    fixed across an NPT trajectory.
 
     Note the heuristic is purely geometric: the resolution PME actually needs also
     grows with the Ewald splitting parameter ``alpha``, so verify the accuracy of a
     chosen mesh against direct Ewald rather than trusting ``spacing`` alone.
     """
-    v = np.asarray(cell_vectors)
-    volume = abs(float(np.linalg.det(v)))
-    lengths = [
-        volume / float(np.linalg.norm(np.cross(v[b], v[c])))
-        for b, c in ((1, 2), (0, 2), (0, 1))
-    ]
+    lengths = np.asarray(cell.perpendicular_lengths).reshape(-1, 3).max(axis=0)
     dims: list[int] = []
     for length in lengths:
         m = int(np.ceil(float(length) / spacing))
@@ -110,6 +112,9 @@ def _bspline_weights(frac: Array, order: int) -> Array:
     """
     j = jnp.arange(order)
     x = frac[..., None] + j  # M_n evaluated at f, f+1, ..., f+(p-1)
+    # M_1 is the indicator of [0, 1), so at this point only j = 0 is nonzero; each
+    # Cox-de Boor pass below widens the support by one grid point, so after the
+    # p - 1 passes all p offsets carry weight.
     w = jnp.where((x >= 0) & (x < 1), 1.0, 0.0)  # M_1
     for k in range(2, order + 1):
         # M_k(x) = x/(k-1) M_{k-1}(x) + (k-x)/(k-1) M_{k-1}(x-1); shift gives M_{k-1}(x-1)
@@ -134,29 +139,28 @@ def _euler_modulus_sq(m_size: int, order: int, dtype: DTypeLike) -> Array:
     return jnp.where(denom2 < 1e-10, 1.0, 1.0 / denom2)
 
 
-def _pme_reciprocal_energy_batched(
-    positions: Array,  # (N, 3) Cartesian, Angstrom
-    charges: Array,  # (N,)
-    sys_idx: Array,  # (N,) per-particle system index
-    n_sys: int,  # number of systems (static)
-    inverse_vectors: Array,  # (n_sys, 3, 3)
-    volume: Array,  # (n_sys,)
-    alpha: Array,  # (n_sys,)
-    mesh: tuple[int, int, int],
-    order: int,
+def _pme_reciprocal_energy(
+    inp: EwaldLongRangeInput[Any], settings: PMESettings
 ) -> Array:
-    """Per-system smooth-PME reciprocal energy, in atomic units (the caller scales
-    by ``TO_STANDARD_UNITS``)."""
+    """Per-system smooth-PME reciprocal energy, shape ``(n_sys,)``, in atomic
+    units (the caller scales by ``TO_STANDARD_UNITS``)."""
+    pc = inp.point_cloud
+    positions = pc.particles.data.positions
+    charges = pc.particles.data.charges
+    sys_idx = pc.particles.data.system.indices
+    n_sys = pc.batch_size
+    cell = pc.systems.data.cell
+    # Key-based lookup (as in `prefactor`), not raw .data: the parameter and
+    # system tables need not be keyed in the same order.
+    alpha = inp.parameters.alpha[pc.systems.index]
+    order = settings.order
     dtype = positions.dtype
-    Mx, My, Mz = mesh
-    Mvec = jnp.asarray(mesh, dtype=dtype)
+    Mx, My, Mz = settings.mesh
+    Mvec = jnp.asarray(settings.mesh, dtype=dtype)
 
-    # Fractional coords in each particle's own cell. The contraction must match
-    # Frame.to_fractional (inverse_vectors contracted on its *first* axis); the
-    # transpose cancels against kvec below, so getting it wrong is invisible for
-    # orthorhombic cells but shears the effective lattice.
-    inv_p = inverse_vectors[sys_idx]  # (N, 3, 3)
-    frac = jnp.einsum("nba,nb->na", inv_p, positions) % 1.0  # (N, 3) in [0, 1)
+    # Fractional coords in each particle's own cell, folded into [0, 1).
+    cell_p = cell[sys_idx]
+    frac, _ = cell_p.fold(cell_p.frame.to_fractional(positions))  # (N, 3)
     u = frac * Mvec
     base = jnp.floor(u).astype(jnp.int32)
     f = u - base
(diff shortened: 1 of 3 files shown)


jonkhler:
* **#196** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#265** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+        volume / float(np.linalg.norm(np.cross(v[b], v[c])))
+        for b, c in ((1, 2), (0, 2), (0, 1))
+    ]
+    dims: list[int] = []
+    for length in lengths:
+        m = int(np.ceil(float(length) / spacing))
+        if multiple_of > 1:
+            m = ((m + multiple_of - 1) // multiple_of) * multiple_of
+        dims.append(max(m, multiple_of))
+    return (dims[0], dims[1], dims[2])
```
n-gao, line 94: This could take a Cell? Since periodicity matters? Also use the Cell methods and also it doesn't support multiple systems and it could run in JAX instead of numpy?

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    those ``None`` keeps the direct-Ewald reciprocal sum.
+
+    Attributes:
+        mesh: Static FFT grid dimensions, held fixed across a trajectory (see
+            `pme_mesh_for_cell`).
+        order: Cardinal B-spline order.
+    """
+
+    mesh: tuple[int, int, int]
+    order: int = DEFAULT_SPLINE_ORDER
```
n-gao, line 69: should this be put into the state? Also looks like it uses the same mesh for all systems?

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    """Order-``p`` cardinal B-spline weights for the ``p`` nearest grid points.
+
+    ``frac`` is the fractional part (``[0, 1)``) of the scaled grid coordinate.
+    Returns ``frac.shape + (order,)`` weights for offsets ``j = 0..order-1`` (grid
+    point ``floor(u) - j``). Weights are smooth in ``frac`` (so gradients flow) and
+    form a partition of unity (sum to 1).
+    """
+    j = jnp.arange(order)
+    x = frac[..., None] + j  # M_n evaluated at f, f+1, ..., f+(p-1)
+    w = jnp.where((x >= 0) & (x < 1), 1.0, 0.0)  # M_1
```
n-gao: this looks weird to me, if frac is between 0 and 1 and I add 0,1,2,3,4,5 then w can only be a 1 for the case where j=0?

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    factor diverges; those modes are damped to 1 rather than left as inf. Even
+    orders (the default) have no such plane.
+    """
+    mvals = _bspline_weights(jnp.zeros((), dtype=dtype), order)  # [M_n(0..p-1)]
+    m = jnp.arange(m_size)
+    k = jnp.arange(order - 1)
+    phase = jnp.exp(2j * jnp.pi * m[:, None] * k[None, :] / m_size)
+    denom = jnp.sum(mvals[1:order][None, :] * phase, axis=-1)
+    denom2 = jnp.abs(denom) ** 2
+    return jnp.where(denom2 < 1e-10, 1.0, 1.0 / denom2)
```
n-gao, line 130: I don't know the algorithm, can you have someone who does also review this?

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    positions: Array,  # (N, 3) Cartesian, Angstrom
+    charges: Array,  # (N,)
+    sys_idx: Array,  # (N,) per-particle system index
+    n_sys: int,  # number of systems (static)
+    inverse_vectors: Array,  # (n_sys, 3, 3)
+    volume: Array,  # (n_sys,)
+    alpha: Array,  # (n_sys,)
+    mesh: tuple[int, int, int],
+    order: int,
+) -> Array:
```
n-gao: use Tables and PMESettings

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    dtype = positions.dtype
+    Mx, My, Mz = mesh
+    Mvec = jnp.asarray(mesh, dtype=dtype)
+
+    # Fractional coords in each particle's own cell. The contraction must match
+    # Frame.to_fractional (inverse_vectors contracted on its *first* axis); the
+    # transpose cancels against kvec below, so getting it wrong is invisible for
+    # orthorhombic cells but shears the effective lattice.
+    inv_p = inverse_vectors[sys_idx]  # (N, 3, 3)
+    frac = jnp.einsum("nba,nb->na", inv_p, positions) % 1.0  # (N, 3) in [0, 1)
```
n-gao: we have Cell methods for that

src/kups/potential/classical/pme.py
n-gao: have someone who knows the algorithm review this

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    probe: Probe[State, Ptch, WithIndices[ParticleId, IsEwaldPointData]] | None = None,
+    gradient_lens: Lens[
+        PointCloud[IsEwaldPointData, HasCell[Periodic3D]], Gradients
+    ] = EMPTY_LENS,
+    hessian_lens: Lens[Gradients, Hessians] = EMPTY_LENS,
+    hessian_idx_view: View[State, Hessians] = EMPTY_LENS,
+    patch_idx_view: View[State, PotentialOut[Gradients, Hessians]] | None = None,
+    *,
+    settings: PMESettings,
+) -> Potential[State, Gradients, Hessians, Ptch]:
```
n-gao, line 233: are incremental updates possible?

test/potential/test_pme.py
```diff
@@ -0,0 +1,218 @@
...
+)
+from kups.potential.classical.pme import (
+    make_pme_long_range_energy,
+    pme_mesh_for_cell,
+)
+from kups.potential.common.graph import PointCloud
+
+from .test_ewald import _make_particle_data, _make_systems
+
+ORDER = 8
```
n-gao, line 39: why a different order than the default above?

src/kups/potential/classical/pme.py
```diff
@@ -0,0 +1,299 @@
...
+    EwaldLongRangeInput,
+    EwaldParameters,
+    IsEwaldPointData,
+    ewald_net_charge_energy,
+)
+from kups.potential.common.energy import PotentialFromEnergy
+from kups.potential.common.graph import PointCloud
+
+DEFAULT_SPLINE_ORDER = 6
+"""Default cardinal B-spline order."""
```
n-gao, line 51: can we infer this from an target epsilon?

test/potential/test_pme.py
```diff
@@ -0,0 +1,218 @@
...
+    """Rebuild an ``EwaldLongRangeInput`` from raw positions + cell matrix (differentiable)."""
+    cell = PeriodicCell(TriclinicFrame.from_matrix(cell_vectors))
+    particles = Table.arange(
+        _make_particle_data(positions, charges, n_systems=1), label=ParticleId
+    )
+    systems = _make_systems(cell[None], cutoff)
+    return EwaldLongRangeInput(PointCloud(particles, systems), params, None)
+
+
+class TestPMEvsEwald:
```
n-gao, line 106: You only test for a very specific regular lattice, can you compare on a random system as well?

n-gao (changes requested):
The code could be more "kUPS", it writes a lot of explicit code for which we have wrappers that include efficient code paths.  
  
I mostly commented on that style since I do not know the algorithm, it would be great to have someone who does review this as well.

src/kups/potential/classical/pme.py
jonkhler: Done in cb56c75 — it now takes a `Cell[Periodic3D]` and uses `cell.perpendicular_lengths` instead of the hand-rolled cross products, and it handles batched cells (per-axis max over systems; exercised in `test_two_system_batch`). It stays host-side numpy deliberately: the output is the static FFT shape, which must be known at trace time, so this is a setup-time helper rather than jittable code (now stated in the docstring).

src/kups/potential/classical/pme.py
jonkhler: The mesh can't live in the traced state — FFT shapes must be static at trace time — so it's construction-time settings, analogous to the fixed k-vector table for direct Ewald. And yes, one mesh shared by all systems: the charge grid is `(n_sys, *mesh)` with one batched FFT over it, and `pme_mesh_for_cell` now sizes it for the largest cell per axis. The `PMESettings` docstring now says both explicitly.

src/kups/potential/classical/pme.py
jonkhler: Correct observation — at that point only j=0 is nonzero, because M_1 is the indicator of [0,1). Each Cox–de Boor pass in the loop below then widens the support by one grid point, so after the p−1 passes all p offsets carry weight (the order-p B-spline has support of width p). Expanded the comment in cb56c75.

src/kups/potential/classical/pme.py
jonkhler: Agreed, will find a domain reviewer. In the meantime the algorithm is pinned against the direct-Ewald oracle: energy, forces, the full dE/dcell tensor, sheared cells, a net-charged cell, a two-system batch, and (new, per the comment below) a fully disordered random system.

src/kups/potential/classical/pme.py
jonkhler: Done in cb56c75 — the kernel now takes `EwaldLongRangeInput` + `PMESettings` directly (with the Table-keyed alpha lookup inside) instead of nine raw arrays.

src/kups/potential/classical/pme.py
jonkhler: Done in cb56c75 — now `cell[sys_idx].frame.to_fractional(positions)` + `Cell.fold` for the [0,1) wrap, replacing the hand-rolled einsum.

src/kups/potential/classical/pme.py
jonkhler: Not usefully: direct Ewald caches per-k structure factors, so one moved particle is a cheap O(N_k) update. In PME a moved particle only touches p³ mesh cells, but the FFT over the whole mesh must be redone regardless — O(M log M) independent of how many particles moved — so there is no cheap incremental path. That's why `cache_lens` is asserted to be `None` here.

test/potential/test_pme.py
jonkhler: Order 8 keeps the spline interpolation error well below the tight tolerances these tests pin — they're chosen to expose convention bugs (transposed lattice, missing background term), not to measure spline accuracy. I checked: at the default order 6 the energy/force comparisons fail at these tolerances. Documented on the constant in cb56c75.

src/kups/potential/classical/pme.py
jonkhler: Yes in principle — Essmann et al. give error estimates that would size mesh and order from (alpha, target epsilon), symmetric to `estimate_ewald_parameters`. It's flagged as future work in `pme_mesh_for_cell`'s docstring; I'd rather do it as a follow-up PR with its own accuracy tests than fold it in here.

test/potential/test_pme.py
jonkhler: Done in cb56c75 — added `test_random_disordered_system_matches_ewald`: 200 uniform-random positions with random neutralized charges in a cubic box, comparing energy and forces against direct Ewald (no lattice symmetry left to mask a convention error).

src/kups/potential/classical/pme.py
jonkhler: Went one level deeper in 8fc4d4e: the P(k) influence function and the units+background tail are now extracted in ewald.py (`reciprocal_prefactor`, `reciprocal_total_energy`) and shared by both paths, so PME structurally cannot drift from direct Ewald on the physics the tests assert they agree on. `make_ewald_long_range_potential` also gained an `energy_fn` parameter, so the PME factory delegates instead of mirroring the composer wiring, and k-vectors/charge scatter now go through `triangular_3x3_matmul` / multi-dim `.at[].add` instead of hand-rolled index math. Net −28 lines.

PR #276, opened 2026-08-15
fix(mcmc): stop double-counting the LJ tail correction in guest stress

### LINKED ISSUE #275: LJ tail correction is counted twice in the reported GCMC/NVT stress
## What broke?

`analyze_mcmc` reports the guest pressure as the sum of three components
(`src/kups/application/mcmc/analysis.py:192`):

```python
all_stress = stress_potential + stress_tail_correction + stress_ideal_gas
```

`stress_potential` already contains part of the Lennard-Jones long-range tail, so the tail is
counted twice — and the two contributions are not even the same quantity.

**Trace.** `make_guest_stress` puts the tail-correction potential into the *differentiated*
potential list (`src/kups/application/simulations/mcmc_rigid.py:490-492`):

```python
potentials = [
    make_lennard_jones_from_state(state_lens, gradient=POSITIONS_AND_CELL),
    make_lennard_jones_tail_correction_from_state(state_lens, gradient=POSITIONS_AND_CELL),
]
```

`global_lennard_jones_tail_correction_energy` reads `cell.volume`
(`src/kups/potential/classical/lennard_jones.py:288`), `POSITIONS_AND_CELL` is the one filter in
`application/potential/filter.py` that does *not* `stop_gradient` the cell, `TriclinicFrame` has
a single pytree leaf `tril` with `volume = |tril[0]*tril[2]*tril[5]|`, and
`PotentialFromEnergy` applies the lens `set` inside the differentiated closure before `jax.vjp`
(`src/kups/potential/common/energy.py:261-273`). So `∂U_tail/∂h = -U_tail·h^-T` flows into
`cell_gradients` and contributes `(U_tail/V)·I` to `stress_potential`. Then
`mcmc_rigid.py:531` adds the closed-form `p_tail` to the `tail_correction` slot.

**Why the two halves are not interchangeable.** Integrating the virial tail integral by parts:

```
P_tail = -dU_tail/dV + (2*pi/3) * sum_ij rho_i rho_j r_c^3 u_ij(r_c)
```

The second term comes from the discontinuity of the truncated pair potential at `r_c` and is
invisible to autodiff. Autodiff through `U_tail` therefore yields only the **cell-derivative contribution**
`-dU_tail/dV`, and that is what gets double counted. (It approaches half of `P_tail` from
above as `r_c >> sigma` -- it is `1.0` at `r_c = sigma` and `0.625` at the potential minimum --
so the boundary term is never a negligible correction.)

## How to reproduce

```python
# Builds ONLY the tail-correction potential with cell gradients enabled.
import jax; jax.config.update("jax_enable_x64", True)
import jax.numpy as jnp
from kups.application.potential.classical.lennard_jones import (
    global_lennard_jones_tail_correction_pressure_from_state,
    make_lennard_jones_tail_correction_from_state,
)
from kups.application.potential.filter import POSITIONS_AND_CELL
from kups.core.cell import PeriodicCell, TriclinicFrame
from kups.core.data.index import Index
from kups.core.data.table import Table
from kups.core.lens import identity_lens
from kups.core.typing import ParticleId, SystemId
from kups.core.utils.jax import dataclass
from kups.potential.classical.lennard_jones import (
    GlobalTailCorrectedLennardJonesParameters,
)

@dataclass
class P:
    positions: jax.Array
    labels: Index[str]
    system: Index[SystemId]

@dataclass
class S:
    cell: PeriodicCell

@dataclass
class St:
    particles: Table[ParticleId, P]
    systems: Table[SystemId, S]

# The ci/statistical NVT case: 50 CO2 (TraPPE) in a 30 A cube, r_c = 12.
L, RC, N_C, N_O = 30.0, 12.0, 50, 100
TRAPPE = {"C_co2": (2.8, 0.0023266791910486473), "O_co2": (3.05, 0.006807690966401598)}
params = GlobalTailCorrectedLennardJonesParameters.from_dict(
    cutoff=RC, parameters=TRAPPE, mixing_rule="lorentz_berthelot", tail_correction=True
)
n = N_C + N_O
key = jax.random.key(0)
particles = Table.arange(
    P(
        jax.random.uniform(key, (n, 3), dtype=float) * L,
        Index.new(["C_co2"] * N_C + ["O_co2"] * N_O),
        Index.new(jnp.zeros(n, dtype=int)),
    ),
    label=ParticleId,
)
cell = PeriodicCell(TriclinicFrame.from_matrix(jnp.eye(3)[None] * L))
state = St(particles, Table.arange(S(cell), label=SystemId))

pot = make_lennard_jones_tail_correction_from_state(
    identity_lens(St), parameters=params, gradient=POSITIONS_AND_CELL
)
out = pot(state).data
g_h = cell.frame.vectors_gradient(out.gradients.cell.data.frame)
Sv = jnp.tril(cell.vectors.mT @ g_h)
sigma = -(Sv + Sv.mT - Sv * jnp.eye(3)) / cell.volume[:, None, None]

print("U_tail / V                ", out.total_energies.data / cell.volume)
print("Tr(sigma)/3 from autodiff ", jnp.trace(sigma, axis1=-2, axis2=-1) / 3)
print("closed-form P_tail        ", global_lennard_jones_tail_correction_pressure_from_state(
    key, state, parameters=params).data)
print("max |dU_tail/dr|          ", jnp.abs(out.gradients.positions.data).max())
```

Output:

```
U_tail / V                 [-5.48809266e-07]
Tr(sigma)/3 from autodiff  [-5.48809266e-07]
closed-form P_tail         [-1.09752785e-06]
max |dU_tail/dr|           0.0
```

The autodiff term is non-zero, equals `U_tail/V` exactly, and is `P_tail` minus the surface term
`-5.487185805182614e-07`. The tail correction has no position gradient at all, so the cell is its
only channel into the stress.

Note the closed-form value `-1.0975278468916126e-06` is **bit-identical** to
`pressure_tail_correction` in `ci/statistical/expected/nvt_50co2_30box.yaml`, which confirms the
reproducer reconstructs that reference case exactly.

## Expected behavior

`stress_potential` should be the configurational virial only, with the tail entering solely
through the `tail_correction` slot. For `ci/statistical/expected/nvt_50co2_30box.yaml` the
spurious contribution is `U_tail/V = -5.48809266373351e-07` eV/A^3 — **3.93 %** of
`pressure_potential` and **1.69 %** of the total pressure:

| field | current | expected |
| --- | --- | --- |
| `pressure_potential` | `-1.3962268401663053e-05` | `-1.3413459135289702e-05` |
| `pressure` | `3.251903782628444e-05` | `3.306784709265779e-05` |
| `pressure_tail_correction` | `-1.0975278468916126e-06` | unchanged |
| `pressure_ideal_gas` | `4.7578834074839093e-05` | unchanged |

Because `exchange_prob: 0` fixes the molecule count and the cell is fixed in NVT, `U_tail/V` is
*constant over the whole trajectory*. So the shift is exactly this constant on the
`stress_potential` diagonal at every step, and **every SEM is unchanged**.

Two further invariants that make the fix checkable:

- The spurious term is **strictly isotropic**, so only the `stress_potential` diagonal moves;
  every off-diagonal entry must stay put.
- `energy`, `loading` and `heat_of_adsorption` must be **bit-identical**. Stress is a pure
  diagnostic, and `stress_fn` is called with a fixed `jax.random.key(0)`
  (`src/kups/application/mcmc/logging.py:161`), so it does not perturb the simulation's RNG
  stream and the trajectory at a given seed is unchanged.

I ran `inputs/ci/nvt_50co2_30box.yaml` with and without the tail correction in the differentiated
potential list (same seed, same machine). All four predictions hold:

```
delta pressure_potential   5.488092663733e-07
predicted -U_tail/V        5.488092663734e-07     relative error 6.4e-15
stress diag shift          [5.48809266e-07 5.48809266e-07 5.48809266e-07]
stress off-diagonal shift  max|.| = 0.000e+00
energy  bit-identical      True
loading bit-identical      True
SEMs                       unchanged to ~1e-21
```

The error grows as rho^2 relative to the ideal rho*kT term, so it is worst at high loading —
exactly the regime adsorption work cares about.

CI does not currently catch this: `validate.py:33` uses a 10-sigma tolerance against the
reference SEM, and the shift is 2.1 sigma.

## Secondary (same function)

`p_tail` is computed on the full `state` while `config_stress` is computed on
`state.guest_only` (`mcmc_rigid.py:526` vs `:531`). `guest_only` sends host rows to the
out-of-bounds sentinel on both the system and species axes, and `InBoundsMask` drops every edge
touching them, so `config_stress` is strictly guest-guest while `p_tail` includes host-host and
host-guest terms. Latent today — `gcmc_co2_rubtak` sets `compute_stress: false`, and the NVT
host `empty_30.cif` is a single dummy `X` with `epsilon = 0`, which the
`tail_corrected = epsilon > 0` mask removes anyway — but it breaks on any real framework.

## Environment

- kUPS: 1.0.4 (`9eb61ea`)
- JAX: 0.10.0
- Python: 3.14.5
- OS: macOS (Apple Silicon, CPU backend)

BekturMur:
Fixes #275.

`make_guest_stress` included the Lennard-Jones tail correction in the list of potentials
differentiated with `gradient=POSITIONS_AND_CELL`. `U_tail` depends on the cell through the
volume, so `dU_tail/dh` landed in `cell_gradients` and added `(U_tail / V) * I` to
`StressResult.potential`. `analyze_mcmc` then sums
`potential + tail_correction + ideal_gas`, where `tail_correction` is the closed-form
`p_tail` — so the cell-derivative part of the tail was counted twice.

The fix removes the tail correction from that list. The closed-form correction stays the
sole tail contribution to the reported stress.

**The energy path is unchanged.** `make_propagator` builds the tail correction with
`gradient=None`; the correction belongs in the energy and stays there.

## Reference values

Only `nvt_50co2_30box` reports stress (`gcmc_co2_rubtak` has `compute_stress: false`). With
`exchange_prob: 0` and a fixed cell, `U_tail / V` is constant over the trajectory, so the
diagonal of `stress` / `stress_potential` shifts by exactly `+5.48809266373351e-07` and the
SEMs are unchanged:

| | before | after |
| --- | --- | --- |
| `pressure_potential` | `-1.3962268401663053e-05` | `-1.3413459135289702e-05` |
| `pressure` | `3.251903782628444e-05` | `3.306784709265779e-05` |

Off-diagonals, all SEMs, and `energy` / `loading` / `heat_of_adsorption` are untouched —
stress is a diagnostic and `stress_fn` runs on a fixed `jax.random.key(0)`, so the trajectory
at a given seed is unchanged.

Note the statistical suite would have passed either way: `validate.py` uses a 10-sigma band
against the reference SEM and the shift is 2.1 sigma. The YAML update is a correctness fix,
not a CI-forced one.

## Test

One regression test added to the existing `TestMakeGuestStress`: toggling the analytical tail
correction must not change `StressResult.potential`, while `tail_correction` itself stays
non-zero. It fails on `main` by `1.99e-05` on the diagonal only.

---

Earlier revisions of this PR bundled unrelated fixes found while investigating (host-vs-guest
tail accounting, and the molecular virial using an unweighted centroid rather than the
documented centre of mass). Those are out of scope here and will follow separately.


src/kups/application/simulations/mcmc_rigid.py
```diff
@@ -485,11 +485,18 @@ def make_guest_stress(
...
+    # where the second term comes from the discontinuity of the truncated pair
+    # potential at ``r_c`` and is unreachable by autodiff. Including the tail
+    # potential here would therefore count the cell-derivative contribution
+    # twice once ``StressResult`` is summed.
     potentials: list[Potential[MCMCState, PositionsAndCell, EmptyType, Any]] = [
         make_lennard_jones_from_state(state_lens, gradient=POSITIONS_AND_CELL),
-        make_lennard_jones_tail_correction_from_state(
-            state_lens, gradient=POSITIONS_AND_CELL
-        ),
     ]
```
n-gao, line 500: Isn't this the only necessary change + updated expected results?

n-gao (commented):
Thanks a lot for spotting this! I agree this seems incorrect. Though I wonder about the size of the PR, isn't the marked code the only necessary change?

src/kups/application/simulations/mcmc_rigid.py
BekturMur: You're right!  That is the only necessary change, and I've cut the PR down to it.

It is now three files:

src/kups/application/simulations/mcmc_rigid.py — the marked removal, plus a short comment so it does not get added back;
ci/statistical/expected/nvt_50co2_30box.yaml — the eight affected reference values;
test/application/test_mcmc_rigid.py — one regression test in the existing TestMakeGuestStress: toggling the analytical tail correction must not move StressResult.potential, while tail_correction itself stays non-zero.

n-gao (approved):
LGTM

src/kups/application/simulations/mcmc_rigid.py
n-gao: Thank you!

PR #277, opened 2026-08-16
feat: add particles_from_arrays constructor

### LINKED ISSUE #273: Direct array-based constructor for Particles and Cell
### Problem or Use Case

kUPS currently provides [`particles_from_ase()`](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/application/utils/particles.py) as the supported application-level path for constructing particle and cell data. Workflows whose data already exists as numerical arrays must first package it into an intermediate `ase.Atoms` object before kUPS can create its JAX-based representation.

For array-native workflows, routing data through an intermediate `ase.Atoms` object requires additional host-side materialization. It also prevents compatible upstream buffers from being passed directly to JAX through mechanisms such as [DLPack](https://arrow.apache.org/docs/python/dlpack.html) and couples external integrations to ASE.

The conversion additionally contains kUPS-specific logic: transforming the cell and Cartesian positions into its lower-triangular frame, determining periodicity, encoding metadata, and constructing `Particles`, `Cell`, `Table`, and `Index` objects. External integrations should not need to reproduce this contract.

### Proposed Solution

Add a public, source-neutral array constructor for particle and cell data representing one complete system. It would accept NumPy-compatible or existing JAX arrays and convert them into the same kUPS representation produced by `particles_from_ase()`.

External adapters could use Arrow-based [ADBC drivers](https://arrow.apache.org/adbc/current/driver/index.html#available-drivers) and [integrations](https://arrow.apache.org/adbc/current/integrations.html) to supply these arrays from databases, dataframe tools, or query engines. These workflows would remain outside kUPS and introduce no ADBC, Arrow, database, or dataframe dependencies.

### Implementation Details

Implement `particles_from_arrays()` alongside `particles_from_ase()`. It would convert inputs to JAX arrays using the active JAX dtype configuration, validate their shapes and field-compatible dtypes without unnecessary host-side copies, transform cell vectors and positions together, and construct the required `Particles`, `Cell`, `Table`, and `Index` objects.

`particles_from_ase()` could reuse shared internal conversion helpers while preserving its current behavior. This would keep the kUPS-specific conversion in one place without imposing the new array contract retroactively on existing ASE inputs.

[MD](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/application/md/data.py) and [relaxation](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/application/relaxation/data.py) state construction could use source-neutral builders internally while retaining the existing ASE-compatible functions as wrappers. MCMC requires additional host, guest, motif, and buffer structures and would remain outside the initial implementation.

Each call would construct one complete system, matching the existing ASE path. Multiple systems could subsequently be combined through kUPS' existing [`Table.union`](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/core/data/table.py) mechanism. External adapters would resolve nulls and assemble streamed chunks before calling the constructor.

### Example API

```python
# Types shown for illustration; exact naming and signature are open for discussion.
def particles_from_arrays(
    *,
    positions: ArrayLike,                    # (n_particles, 3), Å
    cell_vectors: ArrayLike,                 # (3, 3), Å
    periodicity: tuple[bool, bool, bool],
    masses: ArrayLike,                       # (n_particles,), amu
    atomic_numbers: ArrayLike,               # (n_particles,)
    labels: Sequence[str],
    charges: ArrayLike | None = None,        # (n_particles,), e
) -> tuple[
    Table[ParticleId, Particles],
    Cell[AnyPeriodicity],
    Callable[[Array], Array],
]:
    ...
```

The return values would match `particles_from_ase()`: a particle table, the transformed cell with its periodicity, and `uc_transform`, which maps additional Cartesian coordinates into the same lower-triangular frame.

The documented contract should define accepted shapes, units, metadata defaults, and null handling. Positions would be Cartesian and use the same coordinate frame and length units as the supplied cell vectors. A valid cell or bounding frame would be required, including for non-periodic systems. Charges could default to zero when omitted, matching the existing ASE path.

Numeric inputs would follow the active JAX dtype configuration. The constructor would reject incompatible dtype categories but would not impose an additional precision cast.

Tests should verify that the ASE and array paths produce equivalent kUPS representations, particularly for non-orthogonal cells where cell vectors and positions must be transformed together.

### Alternatives Considered

* **Construct existing kUPS types directly.** This requires external integrations to reproduce kUPS-specific coordinate, cell, index, and metadata handling.
* **Accept Arrow objects directly.** This could reduce adapter code but would add an Arrow dependency and bind the public API to one interchange format.
* **Provide database connectors in kUPS.** This would bring querying, schema mapping, null handling, and chunk assembly into the simulation library. A source-neutral array boundary provides broader interoperability with a smaller scope.

sm4rtm4art:
Closes #273.

## Summary

This PR adds a public, source-neutral `particles_from_arrays()` constructor for building kUPS particle and cell data directly from NumPy-compatible inputs or existing JAX arrays. Array-native callers no longer need to create an intermediate `ase.Atoms` object while receiving an equivalent particle table, transformed cell, and coordinate transformation.

## Implementation

* Extracts the shared cell and position transformation and kUPS object construction into `_build_particles_and_cell()`.
* Keeps ASE-specific extraction, charge and label precedence, caching, path loading, and public behavior unchanged.
* Converts numeric inputs independently with `jnp.asarray()` and validates their shapes and dtype categories.
* Uses the active JAX dtype configuration without guaranteeing exact dtype preservation. Integer-typed positions, cell vectors, masses, and charges are promoted with `jnp.result_type(float)`, while atomic numbers retain an integer dtype.
* Defaults omitted charges to floating zeros and normalizes periodicity to an exact Python boolean tuple before cell construction.

## Verification

The added tests cover parity with `particles_from_ase()`, joint position and cell transformation for non-orthogonal cells, returned-transform parity, charge defaults, integer promotion, periodicity dispatch, and invalid shapes, dtypes, labels, and particle-field lengths.

Local pre-commit checks on the changed files, Ruff lint and formatting checks, `pyrefly check src/`, and `git diff --check` all pass. The local full test suite passes with `1908 passed, 7 skipped, 5 warnings`.

## Scope

This change adds no new dependencies and leaves the existing application builders unchanged.

### DIFF
--- a/src/kups/application/utils/particles.py
+++ b/src/kups/application/utils/particles.py
@@ -90,28 +90,23 @@ def _particles_from_path(
     return _particles_from_atoms(next(ase.io.iread(path, index=-1, store_tags=True)))
 
 
-def _particles_from_atoms(
-    atoms: ase.Atoms,
+def _build_particles_and_cell(
+    *,
+    positions: Array,
+    cell_vectors: Array,
+    periodicity: AnyPeriodicity,
+    masses: Array,
+    atomic_numbers: Array,
+    charges: Array,
+    labels: list[Label],
 ) -> tuple[
     Table[ParticleId, Particles], Cell[AnyPeriodicity], Callable[[Array], Array]
 ]:
-    """Build particle data and cell from an ASE Atoms object."""
-    L, uc_transform = to_lower_triangular(jnp.asarray(atoms.cell.array))
-    pbc = (bool(atoms.pbc[0]), bool(atoms.pbc[1]), bool(atoms.pbc[2]))
-    cell = Cell.from_pbc(TriclinicFrame.from_matrix(L), pbc)
-    positions = uc_transform(jnp.asarray(atoms.positions))
-    masses = jnp.asarray(atoms.get_masses())
-    atomic_numbers = jnp.asarray(atoms.get_atomic_numbers())
+    """Build particle data and cell from extracted source data."""
+    L, uc_transform = to_lower_triangular(cell_vectors)
+    cell = Cell.from_pbc(TriclinicFrame.from_matrix(L), periodicity)
+    positions = uc_transform(positions)
     n_atoms = len(masses)
-    charges = jnp.asarray(
-        atoms.info.get(
-            "_atom_type_partial_charge",
-            atoms.info.get("_atom_site_charge", jnp.zeros((len(positions),))),
-        )
-    )
-    labels = list(
-        map(Label, atoms.info.get("_atom_site_label", atoms.get_chemical_symbols()))
-    )
     particles = Table.arange(
         Particles(
             positions=positions,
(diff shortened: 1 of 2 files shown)


n-gao (approved):
LGTM. Thank you for your contribution!

PR #279, opened 2026-08-17
fix(observables): take the molecular virial about the centre of mass

BekturMur:
One of the follow-ups I mentioned on #276. Independent of it — no shared files, no conflict, and
it needs no reference-value changes of its own.

`center_of_mass` gates its mass weighting on `isinstance(particles.data, HasWeights)`, i.e. on a
`weights` attribute — and **no particle type in the package defines one**; they all expose
`masses`. The branch is only ever taken by `test/core/test_position.py`, so in production
"centre of mass" has always meant "unweighted centroid".

**This is not only a reporting issue.** `center_of_mass` also sets the pivot for the rigid-body
rotation move (`mcmc/moves.py`), so it affects sampled trajectories, not just diagnostics.
Rotating about a body-fixed point other than the COM is still valid MC — detailed balance
holds — but it is not the documented convention and it changes the move's behaviour for
asymmetric adsorbates.

Separately, `molecular_stress_via_virial_theorem` documents the RASPA centre-of-mass convention
but carried its own inlined, unweighted copy of the group-centroid computation. It now reuses
`center_of_mass`, which deletes the duplicate.

## Changes

- `center_of_mass` falls back to `HasMasses` when `HasWeights` is absent, leaving the existing
  `HasWeights` contract intact.
- `molecular_stress_via_virial_theorem` takes the reference point from `center_of_mass` instead
  of recomputing it; `IsMolecularVirialParticles` gains `HasMasses`.

## Nothing in the shipped references moves

CO2 is symmetric — C at the origin, two equal O sites at ±1.16 Å — so its centroid and centre of
mass coincide exactly for any masses. Verified bit-identical for a randomly rotated and
translated CO2 pair evaluated with real versus equal masses. The same symmetry means the
rotation pivot is unchanged, so `nvt_50co2_30box` needs no update. Asymmetric or heteronuclear
adsorbates do change, which is the point.

## Tests

`molecular_stress_via_virial_theorem` had no coverage. Three tests, one per property that
delegating the reference point could have broken: equivalence with the atomic formula for
singleton groups, minimum-image handling for a group straddling the boundary, and mass weighting
itself (a heteronuclear pair whose COM at `x = 4.5` and centroid at `x = 3` differ, which fails
on the old code). Plus one test covering the new `HasMasses` branch directly in
`test/core/test_position.py`.

## Question

I took the least invasive option, but the `HasWeights` / `HasMasses` split is your call and I am
happy to redo it: `HasWeights` currently has no implementers anywhere in the package, so it could
equally be removed, or the particle types could grow a `weights` property instead of the fallback
here. Tell me which you prefer.

---

### DIFF
--- a/src/kups/core/utils/position.py
+++ b/src/kups/core/utils/position.py
@@ -14,7 +14,12 @@ from jax import Array
 
 from kups.core.cell import AnyPeriodicity, Cell
 from kups.core.data import Table
-from kups.core.typing import HasPositionsAndGroupIndex, HasWeights, ParticleId
+from kups.core.typing import (
+    HasMasses,
+    HasPositionsAndGroupIndex,
+    HasWeights,
+    ParticleId,
+)
 from kups.core.utils.jax import jit
 from kups.core.utils.segment import segment_sum
 
@@ -36,7 +41,8 @@ def center_of_mass[P: HasPositionsAndGroupIndex](
         yield incorrect results.
 
     Args:
-        particles: Indexed particles, optionally supporting `HasWeights` for mass weighting.
+        particles: Indexed particles, optionally supporting `HasWeights` (or
+            `HasMasses`) for mass weighting; unweighted otherwise.
         cells: Cell(s) carrying lattice geometry and per-axis periodicity.
             Must have one cell per group.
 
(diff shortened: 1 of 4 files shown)


n-gao (approved):
Thanks a lot! LGTM. Would you mind resolving the merge conflict?

PR #286, opened 2026-08-23
feat: add source-neutral MD table builder

### LINKED ISSUE #285: Add source-neutral builder for MD particle and system tables
### Problem or Use Case

[`particles_from_arrays()`](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/application/utils/particles.py), proposed in [#273](https://github.com/cusp-ai-oss/kUPS/issues/273) and added by [#277](https://github.com/cusp-ai-oss/kUPS/pull/277), allows callers to construct kUPS `Particles` and `Cell` data directly from NumPy-compatible inputs, including existing JAX arrays.

The next application-level initialization step remains ASE-specific. [`md_state_from_ase()`](https://github.com/cusp-ai-oss/kUPS/blob/main/src/kups/application/md/data.py) is currently the only provided builder for converting base particle and cell data into the `MDParticles` and `MDSystems` tables used by the MD application.

Array-native callers must therefore create an ASE representation or reproduce kUPS-specific momentum initialization, zero-state construction, unit conversions, and integrator parameters. This particularly affects JAX- or ML-driven workflows that generate structures directly as arrays.

A source-neutral builder would make this path composable without changing simulation orchestration, neighbor-list construction, or potential backends.

### Proposed Solution

Add the following public API to `kups.application.md.data`:
```python
def md_state_from_particles(
    particles: Table[ParticleId, Particles],
    cell: Cell[AnyPeriodicity],
    config: MdParameters,
    *,
    key: Array | None = None,
) -> tuple[
    Table[ParticleId, MDParticles],
    Table[SystemId, MDSystems],
]:
    ...
```

This creates a composable continuation from `particles_from_arrays()` while preserving the public signature and behavior of `md_state_from_ase()`.

### Implementation Details
`md_state_from_particles()` would own the source-neutral MD initialization logic currently embedded in `md_state_from_ase()`. The existing ASE builder would obtain its base data through `particles_from_ase()` and delegate to the new builder.

This is a compatibility-preserving generalization rather than a purely mechanical extraction: the current ASE implementation rebuilds canonical particle and system keys with `Table.arange()`, whereas the new builder would preserve input particle keys and construct the system table from the referenced `SystemId`. The ASE path remains unchanged in observable behavior because `particles_from_ase()` already produces canonical keys.

The public import path would be `kups.application.md.data`. The symbol would be added to that module's `__all__` without adding a package-level re-export.

The name follows the existing `md_state_from_ase()` convention. Both builders return MD particle and system tables rather than a complete `MdState`.

Array-native callers could compose the APIs directly:

```python
particles, cell, _ = particles_from_arrays(...)
md_particles, md_systems = md_state_from_particles(
    particles,
    cell,
    config,
    key=key,
)
```

The builder does not consume the adapter's `uc_transform` because the supplied particles and cell are already expressed in the kUPS coordinate frame.

### Input and Behavior Contract
The builder accepts a non-empty particle table for one complete system and an unbatched cell:
* `particles.data.system.keys` contains exactly one `SystemId`;
* there is one valid system reference per particle, and every reference selects that system;
* `cell.vectors.shape == (3, 3)`;
* `particles` and `cell` already describe the same system in the same kUPS coordinate frame.

The builder does not transform positions or cell vectors. It only adds the system dimension to the cell. Structural violations of this contract raise `ValueError`.

The input particle keys are preserved, and the `MDSystems` table uses the referenced `SystemId` as its key. The builder does not normalize caller-provided IDs.

`Table.union()` remains supported for the canonical zero-based local keys produced by `particles_from_ase()` and `particles_from_arrays()`. Combining independently initialized outputs with arbitrary custom-key layouts is outside the initial union guarantee.

Current MD initialization behavior remains unchanged:
* a provided `key` initializes Maxwell–Boltzmann momenta and removes center-of-mass momentum;
* `key=None` produces zero momenta;
* `MdParameters.initialize_momenta` remains a caller-side orchestration switch and is not interpreted by the builder;
* gradients, cell momentum, potential energy, and default exclusions retain their current initial behavior;
* existing unit conversions, integrator parameters, dtypes, shapes, and PyTree structures are preserved.

Potential construction, neighbor-list sizing, and simulation orchestration remain outside this boundary.

### Scope
* Add the public builder and delegate `md_state_from_ase()` to it.
* Preserve the existing ASE object and file-path workflows.
* Add the public symbol to `kups.application.md.data.__all__`.
* Document the source-neutral input and return contracts.
* Add focused validation, parity, and regression tests.

### Non-Goals
* Constructing a complete `MdState`.
* Accepting multiple systems in one builder call.
* Adding `md_state_from_arrays()` or another ingestion boundary.
* Accepting caller-provided momenta or velocities; the initial builder retains the existing `key`-controlled zero or Maxwell–Boltzmann initialization behavior.
* Adding source-neutral relaxation or MCMC builders.
* Changing the CLI, potential backends, model metadata, neighbor-list logic, or simulation loop.
* Adding new integrator or periodicity validation.
* Adding dependencies, tutorials, benchmarks, or performance guarantees.

### Testing and Verification
Tests should verify that:
1. `particles_from_arrays()` followed by `md_state_from_particles()` is numerically equivalent to the ASE path for the same physical input, configuration, and PRNG key, including an originally skewed cell.
2. `md_state_from_ase()` retains its public signature and both its `ase.Atoms` and file-path behavior.
3. Empty inputs, invalid one-system references, and already-batched cells raise `ValueError`.
4. Custom `ParticleId` keys, the sole `SystemId` key, and the particle-to-system index vocabulary and indices are preserved. ASE-derived outputs retain their canonical key values, while outputs from two canonical source-adapter calls remain compatible with `Table.union()`.
5. `key=None`, fixed-key sampling, center-of-mass momentum removal, and separation from `initialize_momenta` retain their current behavior.
6. Independent builder tests preserve zero-valued fields, default exclusions, base-field dtypes and shapes, and output PyTree structures.
7. Integrator parameters and unit conversions match current behavior for every supported integrator, explicitly including `baoab_npt_langevin`.
8. Focused tests, formatting, linting, type checking, and the full repository test suite pass.

Cross-source floating values should be compared numerically. This change does not add an exact cross-source dtype guarantee.

### Alternatives Considered

* **Normalize keys in the builder.** Canonicalizing IDs would make independently initialized outputs easier to combine, but would discard caller-provided particle and system identities. The proposed builder preserves its input keys and limits the union guarantee instead.
* **Add only `md_state_from_arrays()`.** A direct array builder would duplicate the argument and validation surface of `particles_from_arrays()`. Accepting existing `Particles` and `Cell` values provides a smaller, composable boundary.
* **Construct a complete `MdState`.** A complete state also requires potential-dependent cutoff information and neighbor-list sizing. Keeping those steps in the existing orchestration preserves the separation between source data, MD initialization, and potential configuration.
* **Include relaxation in the same change.** Relaxation uses different cell-filter and deformation-frame initialization semantics and can be evaluated separately after the MD builder pattern is established.

sm4rtm4art:
Closes #285.

## Summary

Adds a public, source-neutral `md_state_from_particles()` builder for constructing one-system `MDParticles` and `MDSystems` tables from existing kUPS `Particles` and `Cell` values.

This allows array-native callers to continue directly from `particles_from_arrays()` without creating an intermediate `ase.Atoms` object. `md_state_from_ase()` retains its public signature and behavior and now delegates MD-specific initialization to the shared builder.

## Implementation

* Preserves the input `ParticleId` keys and the sole referenced `SystemId`.
* Reuses the existing momentum initialization, center-of-mass removal, zero-state construction, unit conversions, and integrator-parameter logic.
* Validates the one-system contract and requires an unbatched cell with `cell.vectors.shape == (3, 3)`.
* Leaves positions and cell vectors unchanged and only adds the system dimension to the cell.
* Uses `key` to select Maxwell–Boltzmann sampling or zero momenta; `MdParameters.initialize_momenta` remains caller-side.
* Exposes the builder from `kups.application.md.data` without adding a package-level re-export.

This is a compatibility-preserving generalization, not only an extraction: the previous ASE path rebuilt canonical keys with `Table.arange()`, while the new builder preserves caller identities. Observable ASE behavior is unchanged because `particles_from_ase()` already emits those canonical keys.

The `uc_transform` returned by source adapters remains available to callers for transforming additional Cartesian data. It is not consumed by this builder because the supplied particles and cell must already share the kUPS coordinate frame.

## Testing

Focused tests cover:

* public signature and export;
* particle and system identity preservation;
* structural validation;
* ASE object and file-path compatibility;
* wrapper delegation;
* skewed-cell array/ASE parity;
* canonical `Table.union()` behavior;
* momentum initialization and center-of-mass removal;
* zero-state values and PyTree structure;
* all five integrators in the implementation base, including the cell-dependent `baoab_npt_langevin` parameters.

The focused and related MD tests, full repository suite, Ruff, Pyrefly, and all configured pre-commit hooks pass.

This change does not construct a complete `MdState` or modify multi-system initialization, relaxation/MCMC builders, CLI orchestration, potential backends, neighbor lists, dependencies, or documentation.

## Merge context

This branch is based on `main` at `01967b6` and does not include [#193](https://github.com/cusp-ai-oss/kUPS/pull/193) or [#194](https://github.com/cusp-ai-oss/kUPS/pull/194). Both currently modify `src/kups/application/md/data.py`, so either landing first may require a mechanical rebase resolution.

The current diffs show no known semantic conflict with the new builder boundary. PR [#193](https://github.com/cusp-ai-oss/kUPS/pull/193) extends integrator construction, which `md_state_from_particles()` reuses through `_build_integrator_params()`; if it lands first, this PR's integrator test matrix would be extended during the rebase to cover the additional integrator.

PR [#194](https://github.com/cusp-ai-oss/kUPS/pull/194) is currently the top of the `#261 → #262 → #263 → #264 → #194` Verlet-skin stack. Its relevant state and orchestration changes remain outside this builder's particle/system initialization responsibility. Merge ordering can therefore be handled through the expected rebase resolution and focused regression tests.


src/kups/application/md/data.py
```diff
@@ -266,38 +266,68 @@ class MdParameters(BaseModel):
     """If True, initialize momenta from Maxwell-Boltzmann distribution."""
 
 
-def md_state_from_ase(
-    atoms: ase.Atoms | str | Path,
+def md_state_from_particles(
```
jonkhler: minor naming nitpick: `md_state_from_particles_and_cell`?

jonkhler (approved):
Thanks a lot for your contribution - that makes sense and looks clean. Minor comment on the naming of the function, but otherwise lgtm!

src/kups/application/md/data.py
sm4rtm4art: Makes sense — I'll rename it to `md_state_from_particles_and_cell`.

Same kind of nit on the tests: keep `test/application/md/test_data.py`, or `test_md_data.py` like `test_mcmc_data.py`?

`test/application/` mixes both layouts.
Happy to follow your preference for this file; a broader tidy feels like a separate issue.

sm4rtm4art: Pushed the rename. 

I'll keep `test/application/md/test_data.py` for now, unless you'd rather change or move it.

jonkhler: that's fine - thanks again! Will merge as soon as the CI runs clear

jonkhler:
Seems to be blocked by CI - can you fix the oustanding issues? Should auto-merge by then.

sm4rtm4art:
> Seems to be blocked by CI - can you fix the oustanding issues? Should auto-merge by then.

Ruff fix is pushed. Auto-merge dropped on the fork push — please re-enable when CI is green.

PR #287, opened 2026-08-24
feat: Add Grimme's D3(BJ) dispersion correction

### LINKED ISSUE #140: Adding D3 dispersion to MLIPs
### Discussed in https://github.com/cusp-ai-oss/kUPS/discussions/125

<div type='discussions-op-text'>

<sup>Originally posted by **stflavian** May 18, 2026</sup>
Hello! While looking though the documentation of kUPS, I could not find a way to add D3 dispersion to MLIPs. Are there any plans to implement this in the future? </div>

BekturMur:
Closes #140.

Adds the pairwise D3 correction with Becke-Johnson damping as a `Potential`
composable with any other through `sum_potentials`, plus a state-binding
adapter, the element reference tables, and a handbook section.

```python
d3 = make_d3_from_state(
    state_lens, parameters=D3Parameters.from_functional("pbe"), gradient=gradient
)
total = sum_potentials(mlip, d3)
```

**Two commits, reviewable separately.** The first (`61f9440`) touches only
`core/potential.py` and its test — it is motivated by D3 composition but is not
part of D3, and is independently green. The second is the dispersion module
itself. Reviewing them in that order keeps 4.6k lines from arriving at once.

## Data provenance and licensing

Read this first — it is the part most likely to need a decision.

Every table comes from an **Apache-2.0** upstream, reproduced numerically
unchanged: `covalent_rad_2009` from `mctc-lib`, and `_r4_over_r2`, `_load_cn`
and `reference-c6.pt` from `tad-dftd3`.

LGPL `simple-dftd3` is used **only** to generate the committed reference values
and is deliberately absent from `pyproject.toml` and `uv.lock`, so no LGPL
package is installed with kUPS or named in its published metadata.
`generate_reference.py` documents how to run it from a throwaway environment.

The numbers ultimately derive from Grimme's GPL-licensed original program; the
Grimme group later released the same data under Apache-2.0 through `mctc-lib`
and `tad-dftd3`, which is the provenance relied on here. If that chain is not
acceptable, `data/` is self-contained and can be replaced without touching the
implementation. Full details in `src/kups/potential/dispersion/data/README.md`.

## Validation

Energies, position gradients and virials are checked against `simple-dftd3`
1.6.0 across **17 systems spanning 13 elements** (116 tests in
`test/potential/dispersion`):

| Group | Systems |
|---|---|
| Molecular | Ar, Ar₂, HF, H₂O, water dimer, CO₂, benzene, benzene dimer |
| Periodic | Si diamond, NaCl, sheared triclinic Si, Si slab (periodic x/y, open z) |
| Transition metals | Cu fcc, Cr bcc, Ni fcc, ZnO wurtzite, Zr hcp |

Tolerances are `1e-9` relative on energy, `5e-9 eV/Å` on gradients and
`5e-10 eV/Å³` on stress — a small factor above the measured worst case, so a
real regression fails rather than hiding under slack.

The metals are chosen to sit in the awkward corners of the C₆ interpolation
rather than to look impressive: `cu_fcc` reaches CN ≈ 8.7 against references
`[0, 0.96]`, so every weight is extrapolated; `cr_bcc` sits above the widest
reference gap in the table; `ni_fcc`'s reference CNs are not sorted; `zno` is
the only cell pairing a metal with a light element, which exercises the
cross-element C₆ block in both index orders.

All 18 D3(BJ) damping parameter sets were diffed against upstream
`simple-dftd3/assets/parameters.toml` and match exactly.

## Scope — what this does not do

- **Two-body term only.** The Axilrod-Teller-Muto three-body term is not
  implemented. It is small for dense molecular solids but not always negligible
  in porous frameworks.
- **Becke-Johnson damping only.** No zero-damping D3(0) variant.
- **The real-space truncation is abrupt** — no shift, no switching function,
  matching `simple-dftd3`, which is what makes the reference agreement exact.
  The step is 2.6 µeV per carbon pair at 15 Å, or order 2 µeV/atom per 0.01 Å of
  motion at condensed-phase density. Small, but it puts a floor on NVE energy
  conservation and on a meaningful force-convergence threshold.
- **No incremental Monte-Carlo path.** The coordination number couples atoms
  beyond the pair being moved, so an old/new edge decomposition would be wrong
  rather than merely slow. `make_d3_from_state` therefore takes no `probe`.
- **13 of 103 elements are numerically covered.** The remaining 90 are not;
  lanthanides and actinides are least exercised.

## Deliberately deferred

Happy to open issues for any of these, or fold them in if you would rather they
land together:

- A MOF case (IRMOF-1) — the metals are covered, a framework is not.
- An Ar₂ binding-curve test, as a physical check that D3 creates a minimum that
  is absent without it.
- A test against a live `mcmc_rigid` state; the current GCMC-padding test builds
  its `Buffered` table directly.
- float32 accuracy. `JAX_ENABLE_X64` is set only under `[tool.pytest_env]`, so
  all validation here is float64 while `docs/simulations.md` offers float32 as a
  supported choice. Measured: ~7e-5 relative on the original periodic cells, up
  to ~1e-4 on ZnO. This is a repo-wide precision policy question rather than a
  D3 one, so I have not changed anything.


docs/handbook.md
```diff
@@ -97,6 +97,46 @@ state = run_md(next(chain), propagator, state, config.run)
...
+
+d3 = make_d3_from_state(
+    state_lens,
+    parameters=D3Parameters.from_functional("pbe"),
+    gradient=POSITIONS_AND_CELL,
+)
+total = sum_potentials(mlip, d3)
+```
+
+That is all that is required. Energies, forces and cell gradients are additive, so stress via [stress_via_virial_theorem][kups.observables.stress.stress_via_virial_theorem] is correct with no extra wiring. Pick the parameterisation with `from_functional` to match whatever functional the MLIP's training data used, or pass `s6`/`s8`/`a1`/`a2` explicitly to `D3Parameters.from_damping`. As with the other force fields, parameters may instead live on the state as `d3_parameters`, which is what you want if they change during a run.
```
jonkhler, line 121: Could we please minimize Claude-speak in the handbook / documentation :)

jonkhler: Specifically minimize fillers like "That is all that is required.". If LLM are used, I recommend to instruct to use terse ASD-STE100 Simplified Technical English as a default.

jonkhler:
Thanks for your addition! Going through the review right now.

General comment: is it possible to break this down in a stack of smaller PRs? 4k+ LOCs usually lead to slow review turnarounds.

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+    D3Parameters,
+    IsD3GraphParticles,
+    make_d3_potential,
+)
+
+__all__ = ["IsD3GraphState", "IsD3State", "make_d3_from_state"]
+
+
+@dataclass
+class _D3ParticleData:
```
jonkhler, line 61: In which way does that differ from the existing ParticleData structs that we have already? 

Also if there is a specific semantics for D3 needed here we should not overload the existing inclusion / exclusion index semantics. If you can elaborate about the intention / difference I can help guide towards the right design.

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@overload
+def make_d3_from_state[State](
+    state: Lens[State, IsD3State],
+    *,
+    parameters: None = None,
+    gradient: None = None,
+    neighborlist_factory: NeighborListFactory[Any] = ...,
+) -> Potential[State, EmptyType, EmptyType, Patch[Any]]: ...
```
jonkhler, line 130: returning `Any` results in invalid types downstream. Can be ok for arguments if type vars would lead to overly verbose solutions. If returned it will break typing downstream.

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@overload
+def make_d3_from_state[State](
+    state: Lens[State, IsD3State],
+    *,
+    parameters: None = None,
+    gradient: Lens[Geometry, PositionsAndCell],
+    neighborlist_factory: NeighborListFactory[Any] = ...,
+) -> Potential[State, PositionsAndCell, EmptyType, Patch[Any]]: ...
```
jonkhler, line 140: Any in return

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@overload
+def make_d3_from_state[State](
+    state: Lens[State, IsD3GraphState],
+    *,
+    parameters: D3Parameters,
+    gradient: None = None,
+    neighborlist_factory: NeighborListFactory[Any] = ...,
+) -> Potential[State, EmptyType, EmptyType, Patch[Any]]: ...
```
jonkhler, line 150: Any in return

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+@overload
+def make_d3_from_state[State](
+    state: Lens[State, IsD3GraphState],
+    *,
+    parameters: D3Parameters,
+    gradient: Lens[Geometry, PositionsAndCell],
+    neighborlist_factory: NeighborListFactory[Any] = ...,
+) -> Potential[State, PositionsAndCell, EmptyType, Patch[Any]]: ...
```
jonkhler, line 160: any in return

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def make_d3_from_state(
+    state: Any,
+    *,
+    parameters: D3Parameters | None = None,
+    gradient: Lens[Geometry, PositionsAndCell] | None = None,
+    neighborlist_factory: NeighborListFactory[Any] = AdaptiveNeighborList.from_state,
+) -> Any:
```
jonkhler, line 169: Should be consistent - either ignore types and let the overloads handle it or use proper return types

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+    state: Lens[State, IsD3GraphState],
+    *,
+    parameters: D3Parameters,
+    gradient: Lens[Geometry, PositionsAndCell],
+    neighborlist_factory: NeighborListFactory[Any] = ...,
+) -> Potential[State, PositionsAndCell, EmptyType, Patch[Any]]: ...
+
+
+def make_d3_from_state(
+    state: Any,
```
jonkhler, line 164: this should always be at least a `IsD3GraphState` as superclass no?

src/kups/application/potential/dispersion/d3.py
```diff
@@ -0,0 +1,233 @@
...
+    gradient_lens: Any = EMPTY_LENS
+    patch_idx_view: Any = None
+    if gradient is not None:
+        gradient_lens = GRAPH_GEOMETRY.nest(gradient)
+        patch_idx_view = position_and_cell_idx_view
+
+    if parameters is not None:
+        param_view: Any = const_lens(parameters)
+    else:
+        param_view = state.focus(lambda x: x.d3_parameters)
```
jonkhler, line 215: untyped `.d3_parameters` field? see above

src/kups/core/potential.py
```diff
@@ -118,14 +115,31 @@ class PotentialOut[Gradients, Hessians]:
     gradients: Gradients
     hessians: Hessians
 
+    def _validate_derivative_structure(
```
jonkhler, line 118: proper typing would not require this validator

src/kups/core/potential.py
```diff
@@ -118,14 +115,31 @@ class PotentialOut[Gradients, Hessians]:
...
+                    f"Cannot {operation} potential outputs with incompatible {label} "
+                    f"structures: left={left}, right={right}. Potentials combined "
+                    "with sum_potentials must use matching gradient and Hessian "
+                    "filters."
+                )
+
     def __add__(
         self, other: PotentialOut[Gradients, Hessians]
     ) -> PotentialOut[Gradients, Hessians]:
+        self._validate_derivative_structure(other, "add")
```
jonkhler, line 136: see above

src/kups/core/potential.py
```diff
@@ -118,14 +115,31 @@ class PotentialOut[Gradients, Hessians]:
...
         return jax.tree.map(jnp.add, self, other)
 
     def __sub__(
         self, other: PotentialOut[Gradients, Hessians]
     ) -> PotentialOut[Gradients, Hessians]:
+        self._validate_derivative_structure(other, "subtract")
```
jonkhler, line 142: see above

src/kups/potential/dispersion/data/__init__.py
jonkhler, line 1: what is the data here?

configuration parameters should come from YAML files - have a look at the existing potentials.

src/kups/potential/dispersion/data/README.md
jonkhler, line 1: why a README in a subfolder? LLM generated?

src/kups/potential/dispersion/data/d3_reference_c6.npz
jonkhler, line 1: we do not store binary data in the repo - what is the need for it?

src/kups/potential/dispersion/d3.py
```diff
@@ -0,0 +1,703 @@
...
+
+D3_DEFAULT_CN_CUTOFF = D3_REFERENCE_CN_CUTOFF
+"""Default coordination-number cutoff [Å], capped at the dispersion cutoff.
+
+``D3Parameters`` takes ``min(cutoff, cn_cutoff)`` so that a single neighbor list
+built at ``cutoff`` always suffices. The counting function decays to a floor of
+``exp(-k1) ≈ 1e-7`` per pair rather than to zero, so the coordination number picks
+up a small, cutoff-dependent tail (of order 1e-3 for a dense solid between 15 Å
+and 21 Å). Match this value to ``simple-dftd3`` when comparing against it.
+"""
```
jonkhler, line 172: these are extensive docstrings on module constants - does it make sense to structure that somehow differently?

src/kups/potential/dispersion/d3.py
```diff
@@ -0,0 +1,703 @@
...
+carry the inclusion/exclusion traits the neighbor list needs.
+
+Separate from [D3Input][kups.potential.dispersion.d3.D3Input] because
+``SumComposer`` is invariant in its input type, so the narrower particle protocol
+the kernels declare is not substitutable for the wider one the graph produces.
+Mirrors the ``LennardJonesInput`` / ``LJRadiusInp`` pair.
+"""
+
+
+def _occupied_atomic_numbers(
```
jonkhler, line 353: why this function with extensive docstring that only unpacks a tuple?

src/kups/potential/dispersion/d3.py
```diff
@@ -0,0 +1,703 @@
...
+    which is not an element. That is exactly the criterion
+    [InBoundsMask][kups.core.neighborlist.masks.InBoundsMask] already uses to
+    drop those rows from the neighbor list, so it is reused here rather than
+    treating ``Z = 0`` as a sentinel in its own right: a genuine ``Z = 0`` in an
+    *occupied* slot is still a bug and must still be caught.
+    """
+    return particles.data.atomic_numbers, particles.data.system.valid_mask
+
+
+def _validate_atomic_numbers_if_concrete(
```
jonkhler, line 370: where is this validation supposed to run?

runtime validations need the kUPS assertion mechanism.

compile time validations should not require a check for a tracer.

src/kups/potential/dispersion/d3.py
```diff
@@ -0,0 +1,703 @@
...
+        validate_atomic_numbers(atomic_numbers)
+    except (
+        jax.errors.ConcretizationTypeError,
+        jax.errors.TracerArrayConversionError,
+    ):
+        # Third-party JAX interpreters may use tracers outside jax.core.Tracer.
+        pass
+
+
+def _assert_supported_atomic_numbers(
```
jonkhler, line 391: so far we handle input validation on the loading / pydantic level. Am I right to understand that this is not something that dynamically changes during the simulation? if so runtime assertions would not make sense here.

src/kups/potential/dispersion/d3.py
```diff
@@ -0,0 +1,703 @@
...
+    Returns:
+        Per-system energies and an identity patch (D3 caches nothing).
+    """
+    _assert_supported_atomic_numbers(*_occupied_atomic_numbers(inp.graph.particles))
+    edge_energy = d3_edge_energy(inp)
+    total_energies = inp.graph.edge_batch_mask.sum_over(edge_energy) / 2
+    return WithPatch(total_energies, IdPatch[Any]())
+
+
+def validate_atomic_numbers(atomic_numbers: Array) -> None:
```
jonkhler, line 595: this explicit validation is again something that we do not do much at other places except for input validation

src/kups/potential/dispersion/damping.py
```diff
@@ -0,0 +1,127 @@
...
+    s6: float
+    s8: float
+    a1: float
+    a2: float
+    doi: str
+
+
+# Fitted values as published in the primary literature cited by each entry.
+# Reproduced with attribution; see ``data/README.md`` for the licensing note.
+BECKE_JOHNSON_PARAMETERS: dict[str, BeckeJohnsonDamping] = {
```
jonkhler, line 63: as before: force-field parameters should be loaded from YAML as in the other examples

src/kups/potential/dispersion/damping.py
```diff
@@ -0,0 +1,127 @@
...
+
+    Returns:
+        The published parameters, with ``a2`` in Bohr.
+
+    Raises:
+        KeyError: If the functional has no tabulated parameters. Pass ``s6``,
+            ``s8``, ``a1`` and ``a2`` explicitly in that case.
+    """
+    key = normalize_functional(functional)
+    try:
```
jonkhler, line 120: using proper pydantic / nanoargs such exceptions would not be required - please have a look

src/kups/application/potential/dispersion/d3.py
jonkhler: I compared it to the reference implementation - there is a subtle but important difference - in the references in the current repo we use a type var `P: Patch[Any]`. Returning `P` is then still propagating the correctly inferred patch type in downstream contexts. Just returning the `Patch[Any]` will always erase type info.

jonkhler (changes requested):
Thanks for your contribution!

This is PR would benefit a lot from:

1. much stronger usage of kUPS idioms / existing code / style (recommend to study our handbook here on the concepts)
2. better separation from force field configuration and implementation
3. moving configurations out of the implementation (the whole `data` subdirectory)
4. strict adherence to our typing standards (within the limits of Python 3.12's typing capabilities we try to not erase types where possible

Also as a further meta-comment: while we use coding agents ourselves, we try to minimize their impact on code that has to be reviewed and read by humans. A proper review of a multi-thousand LOCs PR with lots of AI generated documentation is very difficult. For the sake of turnaround speed and minimizing scope creep accumulating incremental Human-curated additions (e.g. in a PR stack or similar) would help a lot to get this in.

On the content of the PR itself: wouldn't this just boil down to the following components that can be kept minimalist:

1. the implementation of the mathematical per-edge / per-node energy contributions
2. the gift wrapper to make it compatible with the kUPS potential interface
3. the application code + parameters yaml

BekturMur:
Thanks for the review and guidance on splitting this into smaller changes.

I've started replacing this PR with a stack of focused follow-ups. The first one is #296, which contains only the two-body D3(BJ) mathematical kernel and graph energy function.

I'll keep the potential integration and application/configuration layers in separate follow-up PRs.

Closing this PR as superseded by the smaller PR stack.

PR #288, opened 2026-08-25
feat: Add D3 element reference data and Becke-Johnson damping parameters

BekturMur:
First of four pieces splitting #287, as requested there.

Ships the tables Grimme's D3 correction needs, with **no consumer yet**: the element
reference data (covalent radii, `<r⁴>/<r²>` expectation values, reference coordination
numbers, the reference C6 tensor) and the per-functional Becke-Johnson damping
parameters. The kernel that uses them follows in the next PR.

## Provenance and licensing

This is the part most likely to need a decision, so it goes first and stands alone.

Every table comes from an **Apache-2.0** upstream, reproduced numerically unchanged:
`covalent_rad_2009` from [mctc-lib](https://github.com/grimme-lab/mctc-lib), and
`_r4_over_r2`, `_load_cn` and `reference-c6.pt` from
[tad-dftd3](https://github.com/dftd3/tad-dftd3).

The numbers ultimately derive from Grimme's GPL-licensed original DFT-D3 program. The
Grimme group later re-released the same data under Apache-2.0 through those two
projects, and that is the chain relied on here. If it is not acceptable, this
directory is self-contained and can be replaced without touching anything else.

LGPL `simple-dftd3` is used **only** to generate reference values for the test suite
(next PR) and is deliberately absent from `pyproject.toml` and `uv.lock`, so no LGPL
package is installed with kUPS or named in its published metadata.

Full detail, including the sources deliberately not used, is in
`src/kups/potential/dispersion/data/README.md`.

## Units

Tables are stored in whichever unit their own upstream quotes — Ångström for the
covalent radii, atomic units for the rest — so they can be diffed against the source
line by line. `load_d3_reference` is the only place that converts; consumers see
Ångström and electronvolt throughout and never convert again.

The C6 tensor is committed as an upper triangle because the table satisfies
`C6[i, j, a, b] == C6[j, i, b, a]` exactly. Unpacking restores the dense
`(104, 104, 7, 7)` array and keeps the file at ~332 kB, under the
`check-added-large-files` limit.

## Checks

- All 18 damping parameter sets diffed against upstream
  `simple-dftd3/assets/parameters.toml` — exact match, including inherited `s6`
  values and the one entry upstream leaves without a DOI.
- Structural and unit tests in `test/potential/dispersion/test_d3_data.py`: shapes,
  dtypes, the C6 symmetry round-trip, the `4/3` radius scaling, `r4r2` preprocessing,
  and known values such as `C6(Ar,Ar) = 64.6462 au`.
- Full suite green on current `main`.

## Remaining pieces

Will be opened one at a time, each rebased onto `main` after the previous merges, so
every PR carries only its own diff:

2. the D3(BJ) kernel and its validation against `simple-dftd3`,
3. the state-binding adapter and handbook section.

#287 will be closed once this series supersedes it.


jonkhler:
See comments on the `data` folders in the initial PR

PR #292, opened 2026-08-25
refactor: Replace `IsBlockingSpheresProbe.neighborlist` with `particles` and pass probe through blocking-sphere propagators

n-gao:
The blocking spheres potential previously relied on a probe-supplied neighbor list for incremental updates. This approach is replaced so that the probe is now used solely to identify which particles a patch touches, narrowing the sphere query to only those particles rather than sweeping the full configuration. When no probe is provided, the patched state is evaluated over every particle as before.

The `IsBlockingSpheresProbe` protocol is updated to expose a `particles` property (a `WithIndices`-wrapped table) instead of separate `changed_particle_idx` and `neighborlist` properties. The neighbor list is now always constructed from the cutoff factory rather than being sourced from the probe result.

`BlockingSpheresSumComposer.__call__` is restructured so the patch is applied unconditionally when present, and the probe is consulted only to restrict the set of queried particles. The `queried` mask is derived from the probe-narrowed particle set when a probe is available, or from the full valid mask otherwise.

Both `mcmc_rigid` and `mcmc_widom` propagator factories are updated to pass the probe through to `make_blocking_spheres_from_state`, enabling the narrowed evaluation path in those simulations.

Tests are added covering probed and unprobed moves into and outside a blocking sphere, and asserting that the probed path carries only the moved particle while the unprobed path carries the full particle table.

### DIFF
--- a/src/kups/application/potential/classical/blocking.py
+++ b/src/kups/application/potential/classical/blocking.py
@@ -114,8 +114,8 @@ def make_blocking_spheres_from_state(
         state: Lens into the sub-state providing particles, groups, systems, and
             neighbor list (plus ``blocking_spheres_parameters`` when
             ``parameters`` is not given).
-        probe: Probe returning a IsBlockingSpheresProbe; ``None`` for full
-            recomputation.
+        probe: Probe returning a IsBlockingSpheresProbe; ``None`` evaluates a
+            patched state over every particle.
         parameters: Constant blocking sphere parameters. When given they are
             bound with a constant lens and the state need not carry
             ``blocking_spheres_parameters``.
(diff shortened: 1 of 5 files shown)


n-gao:
* **#293** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/293?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#292** <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/292?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/292?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

felixmusil (approved):

n-gao:
### Merge activity

* **Aug 28, 12:22 PM UTC**: @n-gao merged this pull request with [Graphite](https://app.graphite.com/github/pr/cusp-ai-oss/kUPS/292).

PR #296, opened 2026-08-30
feat: add two-body D3(BJ) dispersion kernel

BekturMur:
First small follow-up to #287, limited to the two-body D3(BJ) mathematical kernel and graph energy function.

## Scope

* fractional D3 coordination numbers
* environment-dependent C6 interpolation
* C8 coefficients
* Becke-Johnson pair energy
* per-system energy and coordination cutoffs
* directed-edge reduction
* JAX-safe handling of padded edges

`d3_energy` is the kUPS graph boundary. The lower-level D3 functions operate on arrays.

Application integration, potential construction, parameter YAML, and reference-data loading are intentionally deferred to follow-up PRs.

## Validation

* 13 D3 tests pass
* full test suite passes
* `ruff check` passes
* `ruff format --check` passes
* `pyrefly check src/`: 0 errors

Tests include an independent NumPy evaluation of the D3 pipeline and an Ar dimer energy and gradient comparison against `simple-dftd3`.

### DIFF
--- a/src/kups/potential/dispersion/__init__.py
+++ b/src/kups/potential/dispersion/__init__.py
@@ -1,0 +1,4 @@
+# Copyright 2024-2026 Cusp AI
+# SPDX-License-Identifier: Apache-2.0
+
+"""Dispersion corrections for molecular potentials."""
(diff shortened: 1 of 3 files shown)


BekturMur:
Hi @jonkhler, just a quick ping when you have a chance. I reworked the D3 contribution following your feedback on #287 and split out this first small PR around the mathematical kernel.

Would appreciate your feedback on whether this is now closer to the direction you had in mind.

